PR #4, opened 2017-08-03
feat(lint): add `gts lint` command

JustinBeckwith:


src/lint.ts
```diff
@@ -0,0 +1,33 @@
...
+import * as cp from 'child_process';
+import * as path from 'path';
+import {Options} from './cli';
+
+const gtsRootDir = path.join(__dirname, '../..');
+const tslintPath = path.join(gtsRootDir, '../tslint/bin/tslint');
+
+/**
+ * Run tslint with the default configuration
+ * @param options
```
ofrobots: nit: omit the param annotation since it has no description (preferred), or provide a description.

PR #9, opened 2017-08-03
fix: use local path to module

ofrobots:
The paths we generate in scripts and config should be from the local
node_modules folder rather than pointing to the global directory (may
not persistent as node version changes) or npx directory (may not exist
across invocations).

Also did some refactoring.

Going to start writing tests pretty soon.

### DIFF
--- a/src/init.ts
+++ b/src/init.ts
@@ -24,8 +24,6 @@ const readJson = pify(require('read-package-json'));
 const read = pify(fs.readFile);
 const write = pify(require('write-file-atomic'));
 
-const gtsRootDir = path.join(__dirname, '../..');
-
 interface Bag<T> {
   [script: string]: T;
 }
@@ -45,8 +43,7 @@ async function query(
   return answers.query;
 }
 
-async function generatePackageJson(
-    packageJson: any, options: Options): Promise<void> {
+async function addScripts(packageJson: any, options: Options): Promise<boolean> {
   let edits = false;
   const outDir = 'build/';
   const scripts: Bag<string> = {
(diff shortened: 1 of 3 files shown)


src/init.ts
```diff
@@ -45,8 +43,7 @@ async function query(
   return answers.query;
 }
 
-async function generatePackageJson(
-    packageJson: any, options: Options): Promise<void> {
+async function addScripts(packageJson: any, options: Options): Promise<boolean> {
```
JustinBeckwith: weird nit, but the return type can be implicit, right?

src/init.ts
```diff
@@ -84,8 +81,12 @@ async function generatePackageJson(
       }
     }
   }
+  return edits;
+}
 
-  // Install dev-dependencies.
+async function addDependencies(
```
JustinBeckwith: Same as above, think we shouldn't be explicit about the return type and just let type inference do it's thing

src/init.ts
ofrobots: Consider:

```ts
function doSomthing(): boolean {
 /* forgot to return */
}

function doAnotherThing {
 /* forgot to return */
}

const didSomething  = doSomething();
const didAnotherThing = doAnothingThing();
```

You get a compiler error for `doSomething` but not for `doAnotherThing`. This is an easy mistake to make and I actually made the mistake as I was working on this PR.

I am trying to find a reference in the TS Style guide that mandates that return types are required – for the JS style guide at least [they are required](https://engdoc.corp.google.com/eng/doc/devguide/js/styleguide.md?cl=head#features-functions-returns).

I wonder if tslint can enforce this. It does not seem to be doing so presently.

PR #10, opened 2017-08-03
chore: lint self

ofrobots:


package.json
```diff
@@ -22,11 +22,13 @@
...
     "clean": "rm -rf ./build/",
     "compile": "tsc -p . --rootDir . --outDir build/",
+    "format-check": "echo \"TODO\"",
     "format": "clang-format -i -style='{Language: JavaScript, BasedOnStyle: Google, ColumnLimit: 80}' src/*.ts",
     "lint": "tslint -c tslint.json  --project . --type-check -t codeFrame",
     "lint-fix": "tslint -c tslint.json --project .  --type-check -t codeFrame --fix",
     "release": "standard-version",
-    "test": "echo \"Error: no test specified\" && exit 1"
+    "test": "echo \"TODO\"",
+    "posttest": "npm run lint && npm run format-check"
```
JustinBeckwith: I think I'd rather have linting pretest?

package.json
```diff
@@ -22,11 +22,13 @@
     "build": "npm run compile",
     "clean": "rm -rf ./build/",
     "compile": "tsc -p . --rootDir . --outDir build/",
+    "format-check": "echo \"TODO\"",
```
JustinBeckwith: Do we really need a format check?  Or should we just go ahead and auto-format on build?

ofrobots: This can be dangerous if the user has unsaved edits in their editor. A 'test' job should not make edits IMO.

package.json
ofrobots: For me, in an edit-build-test development cycle I care about lint and format as the very last step once my feature is complete (i.e. I've gotten all the tests to pass).

I find it annouying to be told that my code isn't formatted correctly when I am just trying to see if this is 'functionally correct' w.r.t. tests.

package.json
JustinBeckwith: Got it!

package.json
JustinBeckwith: Ok :)

PR #12, opened 2017-08-03
feat: add `gts fix` command

JustinBeckwith:


src/fix.ts
```diff
@@ -0,0 +1,48 @@
...
+  const gtsRootDir = path.join(__dirname, '../..');
+  const pkgDir = path.relative(options.targetRootDir, gtsRootDir);
+  const tslintPath = path.join(gtsRootDir, '../tslint/bin/tslint');
+  let platform = os.platform();
+  if (platform === 'darwin' || platform === 'linux') {
+    platform += ("_" + os.arch());
+  } else if (platform != "win32") {
+    throw new Error("Fix is only available on the darwin, win32, and linux platforms.");
+  }
+  const clangPath = path.join(gtsRootDir, `../clang-format/bin/${platform}/clang-format`);
```
ofrobots: hmm.. if we were invoked through an npm script, then `clang-format` should already in the path. Your solution is good enough for now, but I would like think upon if there is a way to do this more cleanly.

`npx clang-format` would work, but then we need `npx` to be available....

`$(npm bin)/clang-format` would work except on windows....

src/fix.ts
```diff
@@ -0,0 +1,48 @@
...
+  const lintArgs = ['-c', path.join(pkgDir, 'tslint.json'), '-p',
+                  options.targetRootDir, '-t', 'codeFrame', '--fix'];
+
+  const globby = pify(glob);
+  const srcGlobPath = `${options.targetRootDir}/src/**/*.ts`;
```
ofrobots: nit: add a TODO that we should glob based on the `files`, `includes` and `excludes` section of the `tsconfig.json` file.

src/fix.ts
```diff
@@ -0,0 +1,48 @@
...
+  const srcFiles = await globby(srcGlobPath);
+  const initClangArgs = ['-i', '-style',
+            '{Language: JavaScript, BasedOnStyle: Google, ColumnLimit: 80}'];
+  const clangArgs = initClangArgs.concat(srcFiles);
+  cp.spawnSync(tslintPath, lintArgs, { stdio: "inherit" });
```
ofrobots: Travis should catch this.

src/init.ts
```diff
@@ -49,16 +49,13 @@ async function generatePackageJson(
...
   const scripts: Bag<string> = {
     build: 'npm run compile',
     clean: 'rm -rf ./build/',
     compile: `tsc -p . --rootDir . --outDir ${outDir}`,
     // TODO(ofrobots): use a clang-format wrapper so that it can use the files,
     // include, and exclude sections from tsconfig.json.
-    format:
-        `clang-format -i -style="{Language: JavaScript, BasedOnStyle: Google,` +
-        ` ColumnLimit: 80}" ${defaultSources}`,
+    fix: `gts fix`,
```
ofrobots: nit: Move the todo to `fix.ts`: we should use the `tsconfig.json`'s `files`, `includes` and `excludes` sections to figure out which files should be subject to fix.

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "google-ts-style",
-  "version": "0.1.1",
+  "version": "0.1.9",
```
JustinBeckwith: that was a mistake, rolling back (good catch)

src/fix.ts
JustinBeckwith: Yeah, `npx` would be the ideal way here.  I couldn't really think of another way aside from reaching inside of the module like this (which is also why I had to make it a dep instead of devDep).   Ideas welcomed!

src/fix.ts
JustinBeckwith: Yikes, forgot to run lint

src/fix.ts
JustinBeckwith: fixed

JustinBeckwith:
Addressed all the things!

src/fix.ts
ofrobots: Thing I am worried about.. if `tslint` or `clang-format` gets hoisted to different place in the tree than `google-ts-style`.... I don't have obvious better ideas at the moment. Will continue to do the head-scratching.

PR #15, opened 2017-08-04
feat: add clean command

JustinBeckwith:
Note:  I also went ahead and added the code to load input and output from tsconfig.json here


src/tscfg.ts
```diff
@@ -0,0 +1,56 @@
...
+ * @param tsconfig object containing the ts config
+ */
+export async function getInputFiles(tsconfig: any): Promise<string[]> {
+  const inputFiles: string[] = [];
+  if (tsconfig.files) {
+    Array.prototype.push.apply(inputFiles, tsconfig.files);
+  }
+  if (tsconfig.include) {
+    const options: any = {};
+    if (tsconfig.exclude) options.ignore = tsconfig.exclude;
```
ofrobots: if `tsconfig.exclude` is omitted, it defaults to `['node_modules', 'bower_components', 'jspm_packages']` and ${outDir}.

src/tscfg.ts
```diff
@@ -0,0 +1,56 @@
...
+    for (let globEntry of tsconfig.include) {
+      const globFiles = await globp(globEntry, options);
+      Array.prototype.push.apply(inputFiles, globFiles);
+    }
+  }
+  return inputFiles;
```
ofrobots: Do you want to deal with the case when both `files` and `include` are omitted? (Glob `*.ts[x]` minus excludes.)

I wonder if `tsc` offers a programmatic API to get the list of files it is going to compile?

src/clean.ts
```diff
@@ -0,0 +1,41 @@
...
+import * as tscfg from './tscfg';
+
+const rimrafp = pify(rimraf);
+
+/**
+ * Remove files generated by the build.
+ */
+export async function clean(options: Options): Promise<void> {
+  const tsconfig = await tscfg.getTsConfig(options.targetRootDir);
+  if (tsconfig.compilerOptions && tsconfig.compilerOptions.outDir) {
```
ofrobots: We might want to guard against `outDir` being `.` as well. That's going to be quite disastrous. I guess we require that `outDir` should not have anything that you would like to keep. `.` is going to be a common mistake / misunderstanding.

src/tscfg.ts
ofrobots: Basically, I want the `configParseResult` from [here](https://github.com/Microsoft/TypeScript/blob/8f7a582fc583f7aad98f452790f2ab6c4c560d32/src/compiler/tsc.ts#L257). 

The default expose from the module is exposing a lot of internals. I wouldn't be surprised if we could actually get this implemented this way. What I am not sure about is what the TS team's stance is on semver for everything that is accessible in the module exports.

src/tscfg.ts
JustinBeckwith: awesome catch

src/tscfg.ts
JustinBeckwith: I think adding **/*.ts by default is probably good for now.  Agree that in the future it would be nice to use the TypeScript module to actually do the matching

src/clean.ts
JustinBeckwith: Good call!

PR #33, opened 2017-08-08
lint: start using tslint library api

ofrobots:
https://palantir.github.io/tslint/usage/library/

### DIFF
--- a/src/lint.ts
+++ b/src/lint.ts
@@ -13,19 +13,33 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-import * as cp from 'child_process';
 import * as path from 'path';
+import {Configuration, Linter} from 'tslint';
+
 import {Options} from './cli';
 
 /**
- * Run tslint with the default configuration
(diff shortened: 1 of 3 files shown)


src/lint.ts
```diff
@@ -13,19 +13,33 @@
...
+export function lint(fix: boolean, options: Options): boolean {
+  const tsconfigPath = path.join(options.targetRootDir, 'tsconfig.json');
+  const tslintConfigPath = path.join(options.gtsRootDir, 'tslint.json');
+
+  const program = Linter.createProgram(tsconfigPath);
+  const configuration =
+      Configuration.findConfiguration(tslintConfigPath, '').results;
+  const linter = new Linter({fix: fix, formatter: 'codeFrame'}, program);
+  const files = Linter.getFileNames(program);
+  files.forEach(file => {
```
JustinBeckwith: This will run everything in parallel, right?  Is there a # of files at which this would fall down?

ofrobots: Actually `forEach` is serial.

PR #36, opened 2017-08-09
feat: add format.ts

ofrobots:
Use clang-format API to invoke rather than spawning directly.

Add ability to run check as well as fix.


JustinBeckwith:
LGTM

PR #38, opened 2017-08-11
test: add fixture mechanism for unit testing

ofrobots:
Makes it possible to run runs with fixtures without having out of band
fixtures directories.

Add tests for lint.ts.

### DIFF
--- a/test/fixtures/kitchen/tsconfig.json
+++ b/test/fixtures/kitchen/tsconfig.json
@@ -1,5 +1,5 @@
 {
-  "extends": "./tsconfig-google.json",
+  "extends": "./node_modules/google-ts-style/tsconfig-google.json",
   "compilerOptions": {
     "rootDir": ".",
     "outDir": "build"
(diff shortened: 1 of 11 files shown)


JustinBeckwith:
This is awesome!

ofrobots:
Tests are green. PTAL.

ofrobots:
Imma going to land this in the interest of progress. Please open issues if you have comments.

test/fixtures/kitchen/tsconfig.json
```diff
@@ -1,5 +1,5 @@
 {
-  "extends": "./tsconfig-google.json",
+  "extends": "./node_modules/google-ts-style/tsconfig-google.json",
```
matthewloring: Will these paths always be relative to the project root?

ofrobots: This is a system test. It installs ts-style locally. For this system test, this is the correct system path. Did you have a more specific question?

Note that this system test needs to be expanded a lot more. It should go through the exercise of creating a temp directory out-of-tree (so that we don't inherit any node_modules from the parent folder), running `google-ts-style init`, dropping in some TS code and then making sure everything is setup correctly for the project.

Issue #39, opened 2017-08-11
Allow leading underscores in variable names


murgatroid99:
The tsconfig file provided here includes the `noUnusedParameters` compiler option, which is useful. However, sometimes when implementing an interface, some parameters are not actually useful. In that case, if the argument name starts with an underscore, the compiler will not emit an error (see [this documentation](https://github.com/Microsoft/TypeScript/wiki/What%27s-new-in-TypeScript#flag-unused-declarations-with---nounusedparameters-and---nounusedlocals)). Currently, if I try to do this, then I try to lint my files using the TSLint configuration provided here, it fails because the lint configuration does not allow any underscores in non-constant names.

Please fix this by adding the `"allow-leading-underscore"` option to the TSLint `"variable-name"` rule configuration.

ofrobots:
Summoning @mprobst as this is a style guide issue.

(Hi @mprobst! this repo isn't quite ready yet, but it would be good to have co-ownership of this project between the typescript teams and the node team).

mprobst:
Internally at Google we do not use the `noUnusedParameters` compiler flag, among other things for this reason. We find that `noUnunsedParameters` is super useful as a presubmit-style check, but should not fail compilation. It's in our experience annoying in an actual developer workflow, so it should be a warning, and only trigger later in your workflow.

Because of that, we're instead relying on tslint's `no-unused-variable` check. That gets you most of the value, though it doesn't (IIRC) have complete support for parameters. But that could be added, I presume.

I'd rather not allow underscores in variable names, as you're kind of throwing the baby out with the bathwater here - you're addressing a limitation in the tooling by opening up a much bigger limitation, which is not enforcing variable names properly across the entire code base, instead of just specifically handling parameters. I'd personally value enforcing reasonable, consistent naming conventions higher than using `noUnunsedParameters`, given that I find it's a bit of an anti-pattern in the workflow and has a decent replacement with tslint's `no-unused-variable` check.

ofrobots:
Hmm.. I wonder where we inherited the `noUnusedParameters` compile flags. The goal was to try and mirror the internal config. I will go and re-audit the list.

I agree with your argument on `noUnusedParameters`.

ofrobots:
PR: https://github.com/google/ts-style/pull/40

PR #40, opened 2017-08-14
drop noUnusedParameters

ofrobots:
The internal Google configuration doesn't use this flag.
Fixes: https://github.com/google/ts-style/issues/39


tsconfig-google.json
```diff
@@ -10,7 +10,6 @@
     "noEmitOnError": true,
     "noImplicitReturns": true,
     "noUnusedLocals": true,
```
mprobst: we also don't include this.

ofrobots: Thanks. Dropped `noUnusedLocals` as well.

Issue #35, opened 2017-08-09
discussion: replace 'lint' with 'check'


ofrobots:
We right now have the following lint / format related rules:
* `lint`: runs linter, prints errors
* `fix`: runs linter, runs formatter, automatically fixes problems.

We are missing a verb to check format. Instead of adding a 'format' verb, I propose that we rename `lint` to `check` with the semantics that it would check both lint and format problems.

PR #43, opened 2017-08-16
breaking: change default verbs + cleanup

ofrobots:
* Replace `lint` with `check` which checks both format and lint errors.
* Remove unnecessary `build` script.
* Add prepare, pretest and posttest scripts to target.
* Terminate with proper exit code on failures
* Get rid of trivial fix.ts
* README improvements.

Fixes: https://github.com/google/ts-style/issues/35


src/cli.ts
```diff
@@ -44,7 +44,7 @@ const cli = meow(`
     --dry-run     Don't make any acutal changes.
 
 	Examples
-    $ gts init
+    $ gts init -y
```
JustinBeckwith: is it still gts, or google-ts-style?

ofrobots: naming is still unresolved :(

Issue #48, opened 2017-08-18
"Cannot find module 'typescript'" when running `npx google-ts-style@0.2 init`


kjin:
Despite installing `typescript`, the module requires `typescript` to be installed locally first. I think this wasn't the behavior in 0.1.2.

It seems like some cases, this might not work even with locally installed `typescript`.

ofrobots:
Fixed in #49.

PR #49, opened 2017-08-18
fix: conditionally require lint, format

kjin:
Addresses #48

### DIFF
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -33,8 +33,8 @@ export interface Options {
   logger: Logger;
 }
 
-export type VerbFunction =
-  (options: Options, fix?: boolean) => Promise<boolean>;
+export type VerbFunction = (options: Options, fix?: boolean) =>
+    Promise<boolean>;
 
 const logger: Logger = console;
 
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -17,7 +17,6 @@
 import * as meow from 'meow';
 import * as updateNotifier from 'update-notifier';
 import {init} from './init';
-import {format} from './format';
 import {clean} from './clean';
 
 export interface Logger {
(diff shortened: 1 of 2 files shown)


ofrobots:
@JustinBeckwith WDYT?

If we are doing this, then we should also add a `compile` verb, and stop requiring the user to depend on `typescript`. Otherwise the user may have a different version of typescript than the one we use and may see spurious errors with one but not the other. Is there any reason other than the `compile` verb for the user to depend on typescript?

This would imply that typescript version is determined by `gts` rather than the user – who get locked into the version we use.

Another way to fix this would be to make sure we never have import `typescript` on the path reachable from `gts init`. @kjin do you this this is feasible?

JustinBeckwith:
I appreciate what you're trying to do - but I think this introduces more problems than it solves.  It's so very important to let the user control the version of `tsc` they use with their project.  That's truly going to be different for every user. 

Is there a way to have init fail if typescript isn't already installed?  Or install typescript as a peer dependency for the user?

kjin:
@ofrobots @JustinBeckwith PTAL

src/cli.ts
```diff
@@ -74,9 +75,13 @@ async function run(verb: string): Promise<boolean> {
     yes: cli.flags.yes || cli.flags.y || false,
     logger: logger
   };
+  // See: https://github.com/google/ts-style/issues/48
```
ofrobots: nit: it would be better to explain the reason why we are doing this here in a comment. You can additionally link to the issue, but the gist of the intention should be clear here.

ofrobots:
For posterity, it is worth documenting why we don't want to directly depend on typescript (i.e. enforce the typescript version for the users of `gts`):

* TypeScript [doesn't follow semver](https://github.com/Microsoft/TypeScript/issues/14116). This means that a minor update to the typescript version could result in compile errors in users' code. Directly depending on `typescript` would mean breaking changes for `gts` users whenerver we bump the typescript dependency.
* Users may want newer features so they may want the latest version of TypeScript. Each user should be able to move at their own pace on when they want to adopt the new features. They need to be at the minimum enforced by the `peerDependency` requirement we have, but if they want to be newer, that is fine.

PR #47, opened 2017-08-18
breaking: rename module

ofrobots:
* Use `gts` as the only name going forward.
* Ability to check format.

PR #51, opened 2017-08-30
Update the `npm install` command in the README

DominicKramer:


ofrobots:
The original was correct. See: https://github.com/google/ts-style/pull/47. It is just that we haven't published the new `gts` module yet. Waiting for https://github.com/google/ts-style/pull/49 to land.

PR #52, opened 2017-09-08
fix: fix broken commands

kjin:
#49 broke `gts check` (and `gts fix`) by omitting a `module.exports` in `lint.ts`.

In addition to fixing the above, I've added tests for this, and run `npm run lint`.

### DIFF
--- a/package.json
+++ b/package.json
@@ -26,7 +26,7 @@
     "format": "clang-format -i -style='{Language: JavaScript, BasedOnStyle: Google, ColumnLimit: 80}' src/*.ts test/*.ts",
     "lint": "tslint -c tslint.json  --project . --type-check -t codeFrame",
     "lint-fix": "tslint -c tslint.json --project .  --type-check -t codeFrame --fix",
-    "prepublish": "npm run compile",
+    "prepare": "npm run compile",
     "test": "npm run compile && ava build/test/test*.js",
     "posttest": "npm run lint && npm run format-check"
   },
--- a/package.json
+++ b/package.json
@@ -26,6 +26,7 @@
     "format": "clang-format -i -style='{Language: JavaScript, BasedOnStyle: Google, ColumnLimit: 80}' src/*.ts test/*.ts",
     "lint": "tslint -c tslint.json  --project . --type-check -t codeFrame",
     "lint-fix": "tslint -c tslint.json --project .  --type-check -t codeFrame --fix",
+    "prepublish": "npm run compile",
     "test": "npm run compile && ava build/test/test*.js",
     "posttest": "npm run lint && npm run format-check"
   },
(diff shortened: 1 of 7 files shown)


package.json
```diff
@@ -26,6 +26,7 @@
     "format": "clang-format -i -style='{Language: JavaScript, BasedOnStyle: Google, ColumnLimit: 80}' src/*.ts test/*.ts",
     "lint": "tslint -c tslint.json  --project . --type-check -t codeFrame",
     "lint-fix": "tslint -c tslint.json --project .  --type-check -t codeFrame --fix",
+    "prepublish": "npm run compile",
```
ofrobots: I think `prepare` is perferred over `prepublish`, but that might be a seasonal thing (too lazy to find citation).

src/lint.ts
```diff
@@ -48,3 +48,8 @@ export function createProgram(options: Options): ts.Program {
   const tsconfigPath = path.join(options.targetRootDir, 'tsconfig.json');
   return Linter.createProgram(tsconfigPath);
 }
+
+module.exports = {
```
ofrobots: Why is this needed? I think `lint` and `createProgram` are already exported, so the `require('./lint).lint` in `cli.ts` should work without this, unless I am mistaken.

package.json
kjin: Yes, you're right -- changed

src/lint.ts
kjin: Done -- seems like the root problem was simply omitting the `.lint` in the first place.

src/util.ts
```diff
@@ -37,7 +37,7 @@ export function nop() {
 export async function getTSConfig(
     rootDir: string, parmReadFilep?: ReadFileP): Promise<any> {
   const tsconfigPath = path.join(rootDir, 'tsconfig.json');
-  const json = await (parmReadFilep || readFilep)(tsconfigPath, 'utf8');
+  const json = await(parmReadFilep || readFilep)(tsconfigPath, 'utf8');
```
ofrobots: wat? did the formatter produce this? Seems wrong. Nothing you can address though.

kjin: Yep -- likely a bug in the formatter. I've modified the code a little to prevent this.

src/util.ts
```diff
@@ -35,9 +35,12 @@ export function nop() {
  * @param rootDir Directory where the tsconfig.json should be found.
  */
 export async function getTSConfig(
-    rootDir: string, parmReadFilep?: ReadFileP): Promise<any> {
+    rootDir: string, paramReadFilep?: ReadFileP): Promise<any> {
   const tsconfigPath = path.join(rootDir, 'tsconfig.json');
-  const json = await(parmReadFilep || readFilep)(tsconfigPath, 'utf8');
+  if (!paramReadFilep) {
```
ofrobots: nit: more idiomatic to do `paramReadFilep = paramReadfilep || readFilep`. Can we rename the variable because it no longer contains just the parameter value though.

kjin: I named it `customReadFilep`.

PR #55, opened 2017-09-15
Replace read-package-json with vanilla JSON.parse

kjin:
`read-package-json` [inserts a number of fields](https://github.com/GoogleCloudPlatform/cloud-trace-nodejs/pull/554/files/6e59f889f953868ebe198969be7908400af202f1..e875441c58e0de2cfd0e1784605cfcf0d6986311#r139029874) that we likely don't want to write to the user's `package.json` file. This change replaces it with `readFilep` followed by `JSON.parse`.

We might lose some guarantee about the structure of the `package.json` file being read (seems like `read-package-json` does some extra checks)... but these cases might be unrecoverable either way.

### DIFF
--- a/src/init.ts
+++ b/src/init.ts
@@ -25,10 +25,6 @@ interface Bag<T> {
   [script: string]: T;
 }
 
-function noop() {
-  /* empty */
-}
-
 async function query(
     message: string, question: string, defaultVal: boolean,
     options: Options): Promise<boolean> {
@@ -184,7 +180,7 @@ async function generateTsConfig(options: Options): Promise<void> {
 export async function init(options: Options): Promise<boolean> {
   let packageJson;
   try {
-    packageJson = await readJson('./package.json', noop, true /*strict*/);
+    packageJson = await readJson('./package.json');
   } catch (err) {
     if (err.code !== 'ENOENT') {
       throw new Error(`Unable to open package.json file: ${err.message}`);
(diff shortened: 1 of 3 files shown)


src/init.ts
```diff
@@ -184,7 +180,7 @@ async function generateTsConfig(options: Options): Promise<void> {
 export async function init(options: Options): Promise<boolean> {
   let packageJson;
   try {
-    packageJson = await readJson('./package.json', noop, true /*strict*/);
+    packageJson = await readJson('./package.json');
```
ofrobots: Would `require('./package.json')` work?

kjin: I don't think `require` will work since it loads a file relative to `build/src/init.js`.

ofrobots: How about `require(path.join(process.cwd(), 'package.json'))`?

kjin: That would work, but in my opinion we should prefer explicitly reading a file over using a function that is meant to load modules. In this case, we are reading a JSON file, modifying its contents, and then overwriting it. Even though it's a totally valid operation to `require` a JSON-file, I think its use should be restricted to files that are treated as data-only modules (such as a constants file).

ofrobots: It is not quite accurate to say that `require` is meant to load modules. Node core fully supports load JSON files.

If we should have a `readJSON` function, we need to be able to articulate how `require` is inadequate. You have stated preferences, but not a rationale.

kjin: `require` is by no means inadequate, but I would still argue that it is meant primarily to load modules. My rationale is that there's a distinction between "module" and "data" JSON files -- "module" JSON files being portable/shorthand form for JS modules that only export static values (and therefore part of the application's source code), and "data" JSON files being among the application's I/O. That they happen to have the same file extension and data format should be irrelevant. I would consider reading and writing the user's `package.json` file to be the latter.

ofrobots: I don't agree, but this was a nit, and if you feel strongly about this, I am not going to block.

PR #61, opened 2017-10-13
support running gts on individual files instead of the whole project

jinwoo:


ofrobots:
Although it would be good to update the commit message to be <50 chars and to use a angular style prefix for the type. E.g. 'feat: support ...'

jinwoo:
Updated the commit message.

ofrobots:
Feel free to land once the CI is green.

Issue #60, opened 2017-10-13
`gts check` can fail with no output


kjin:

KamranAsif:
This is blocking me as well. Any updates on this?

kjin:
A fix for this should be landed shortly (#62); all it does is add an error message if linting succeeds and formatting fails. I'm planning to add parsing `clang-format` XML output as a follow-up (PRs from your end would be more than welcome too). That being said, it's best to run `gts fix` to fix format issues anyway; I believe `clang-format` output is not necessarily meant to be human-interpretable.

PR #62, opened 2017-10-13
feat: warn when clang-format fails

kjin:


ofrobots:
Odd that travis wasn't launched for this.

kjin:
@ofrobots Probably because it's unmergeable... rebased

DominicKramer:
LGTM: provided the tests pass

ofrobots:
When landing, make sure the commit includes the 'Fixes: #60' metadata.

kjin:
Since the approvals were issued, I added a test... was able to get code coverage to increase by [0.01%](https://codecov.io/gh/google/ts-style/pull/62)

A point of possible contention is [here](https://github.com/google/ts-style/pull/62/commits/f732709963e93f92434b569e2048ccb18a419d6c#diff-887ef50994f94df2ed539a9bdcb6a8e8R25). The explanation is that `exec` shouldn't actually be promisified the way it is because the callback may populate `err` _and_ `stdout`/`stderr`. So I added a new function to always resolve with an exit code (though most places still use the now-renamed "simple" `execp`)

Also, two things happened:
* `clang-format` updated, meaning things had to be re-formatted
* `chalk` got built-in TS definitions that clash with `@types/chalk`, so I removed the dependency on the latter (2e4711a)

@DominicKramer @ofrobots PTAL at the two linked things.

test/test-kitchen.ts
```diff
@@ -8,14 +8,31 @@ import * as pify from 'pify';
...
 const pkg = require('../../package.json');
 
 const rimrafp = pify(rimraf);
 const mkdirp = pify(fs.mkdir);
-const execp = pify(cp.exec);
+const simpleExecp = pify(cp.exec);
 const renamep = pify(fs.rename);
 const ncpp = pify(ncp.ncp);
 
+const execp =
```
ofrobots: This deserves a comment. Also, we should probably look for an ecosystem module for execp.

ofrobots:
Rule of thumb: if you have explain why some code is the way it is in the PR, you probably also need it as a comment.

LGTM once that is addressed.

PR #72, opened 2017-11-01
chore: restrict typescript version

ofrobots:
TypeScript doesn't follow semver. The 2.6 release today is breaking our
CI. This change locks it down to a safer semver range.


JustinBeckwith:
LGTM

PR #74, opened 2017-11-02
fix: add support for typescript 2.6

ofrobots:


package.json
```diff
@@ -66,7 +66,7 @@
     "nyc": "^11.2.1",
     "source-map-support": "^0.5.0",
     "tmp": "0.0.31",
-    "typescript": "~2.4.1"
+    "typescript": "^2.6.1"
```
JustinBeckwith: Do we want to go with the ~ here?

test/test-kitchen.ts
```diff
@@ -22,18 +22,33 @@ const simpleExecp = pify(cp.exec);
 const renamep = pify(fs.rename);
 const ncpp = pify(ncp.ncp);
 
+// TODO: improve the typedefinitions in @types/node. Right now they specify
+// the return type to be Error.
+interface ExecError extends Error {
```
JustinBeckwith: This is non-blocking PR commentary :)  I see the pattern of attaching a `code` property to Error objects all over our repositories.  It makes working with TypeScript and errors kind of ugly.  Two thoughts on how to address:

1. Should we maybe consider using [Error.prototype.name](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error/name) instead?

2. Should we consider chatting with tc39 folks about adding a general purpose `code` field to the Error object?

ofrobots: @jinwoo No. `ErrnoException` has a string `code`. `child_process.exec` API returns an `Error` with a number `code` property. `@types/node` doesn't have an error type for the `exec` at the moment.

package.json
ofrobots: Thanks for catching that!

test/test-kitchen.ts
ofrobots: @JustinBeckwith 

> Should we maybe consider using Error.prototype.name instead?

This is not up to us. The `exec` API in Node.js core returns an `Error` with a particular structure.

> Should we consider chatting with tc39 folks about adding a general purpose code field to the Error object?

In Node.js core, James Snell is championing an effort to have a `code` property added to all errors returned by core. I think it would be worth chatting with him to the next steps beyond that.

ofrobots:
TBR= @kjin

test/test-kitchen.ts
```diff
@@ -22,18 +22,33 @@ const simpleExecp = pify(cp.exec);
...
+  code: number;
+}
+
+function isExecError(err: Error|ExecError): err is ExecError {
+  return (<ExecError>err).code !== undefined;
```
kjin: I think we prefer `err as ExecError`?

test/test-kitchen.ts
```diff
@@ -22,18 +22,33 @@ const simpleExecp = pify(cp.exec);
...
 // code, so we define a more exhaustive promsified cp.exec here.
+// TODO: replace this code with a npm modules that promisifies exec.
 const execp =
     (command: string, execOptions?: cp.ExecOptions): Promise<ExecResult> => {
       return new Promise((resolve) => {
         cp.exec(
             command, execOptions || {},
-            (err: Error&{code: number}, stdout, stderr) => {
-              resolve({exitCode: err ? err.code : 0, stdout, stderr});
+            (err: Error|ExecError|null, stdout, stderr) => {
```
kjin: Should this be `ExecError|null`? This looks like the same as https://github.com/stephenplusplus/gcp-metadata/pull/16

If so, I think `isExecError` might not be needed.

test/test-kitchen.ts
kjin: Also, `isExecError` could have checked that `err` is undefined.

PR #75, opened 2017-11-06
BREAKING CHANGE: chore: sync with internal lint

ofrobots:
This synchronizes the lint rules to be closer to the lint rules used
internally at Google. Not all rules are available externally, so this
is not *exactly* the same.

Also updated the TypeScript version to 2.6.1.

A fair number of new lint rules are being applied now, which means that
the code will be alot more consistent with other Google code, but it
also means that users of gts may need to modify the source to pick up
this release. For most of the new lint rules, a fixer is also available
which should help with migration.

TypeScript 2.6.x enables contravariant types for function parameters,
which may also require some changes. See [1] for more info.

[1] https://blogs.msdn.microsoft.com/typescript/2017/10/31/announcing-typescript-2-6/


tslint.json
```diff
@@ -1,31 +1,64 @@
...
   "rules": {
+    "array-type": [true, "array-simple"],
+    "arrow-return-shorthand": true,
+    "ban": [true,
+      {"name": "parseInt", "message": "tsstyle#type-coercion"},
+      {"name": "parseFloat", "message": "tsstyle#type-coercion"},
+      {"name": "Array", "message": "tsstyle#array-constructor"}
+    ],
+    "ban-types": [true,
+      ["Object", "Use {} instead."],
```
jinwoo: What's the rationale for this rule? Just curious.

tslint.json
```diff
@@ -1,31 +1,64 @@
...
+    "use-isnan": true,
+    "variable-name": [
+      true,
+      "check-format",
+      "ban-keywords",
+      "allow-leading-underscore",
+      "allow-trailing-underscore"
+    ]
   }
 }
```
jinwoo: Nit. newline?

tslint.json
kjin: +1, as a reader I prefer `Object`

ofrobots: https://www.typescriptlang.org/docs/handbook/declaration-files/do-s-and-don-ts.html

ofrobots: More: https://stackoverflow.com/questions/18961203/typescript-any-vs-object/28795689#28795689

PR #77, opened 2017-11-08
chore: drop some lint rules

ofrobots:
This brings us further in sync with the internal lint rules. This bring
us further in sync with the internal tslint rules.

We still have the following extra rules:
* no-arg: disallows arguments.callee
* no-conditional-assignment: disallows any type of assignment in
  conditionals.
* no-duplicate-variable: disallows duplicate var in same block.


jinwoo:
Are they not needed because we're using clang-format?

ofrobots:
The whitespace changes can be dropped because clang-format ensures all the format rules. Some other rules are being dropped because they are not present internally, and I don't have a very strong reason why we *must* continue to have them (i.e. I can see valid uses on the contrary). The three that I didn't drop are because I don't see a valid reason to use `arguments.callee` in Node.js code for example. I plan to investigate why these aren't present internally.

PR #84, opened 2017-11-13
docs: improve readme

JustinBeckwith:


README.md
```diff
@@ -35,6 +42,13 @@ $(npm bin)/gts init
   - `compile`: Compiles the source code using TypeScript compiler.
   - `pretest`, `posttest` and `prepare`: convenience integrations.
 
+## Individual files
+The commands above will all run in the scope of the current folder.  Some commands can be run on individual files:
+
+```js
+$ gts check index.ts
```
jinwoo: Multiple files can be passed as well, so it'd be good to have that case as an example. E.g.
```shell
$ gts check index.ts another.ts ...
$ gts check *.ts
```

ofrobots:
Nit: use doc: prefix instead of chore. LGTM w/ @jinwoo suggestion

README.md
JustinBeckwith: Ack - updated!

Issue #86, opened 2017-11-13
`gts check` leads to `TypeError: Cannot read property 'getFullText' of undefined`


JustinBeckwith:
```
beckwith-macbookpro:google-api-nodejs-client beckwith$ ls
AUTHORS             COPYING             apis                lib                 samples             test                yarn.lock
CHANGELOG.md        MIGRATING.md        coverage            node_modules        scripts             tsconfig.json
CONTRIBUTORS        README.md           jsdoc-conf.json     package.json        templates           tsconfig.tools.json
beckwith-macbookpro:google-api-nodejs-client beckwith$ gts check lib/*.ts
TypeError: Cannot read property 'getFullText' of undefined
    at /Users/beckwith/.nvm/versions/node/v8.9.1/lib/node_modules/gts/build/src/lint.js:35:55
    at Array.forEach (<anonymous>)
    at lint (/Users/beckwith/.nvm/versions/node/v8.9.1/lib/node_modules/gts/build/src/lint.js:34:14)
    at /Users/beckwith/.nvm/versions/node/v8.9.1/lib/node_modules/gts/build/src/cli.js:77:46
    at step (/Users/beckwith/.nvm/versions/node/v8.9.1/lib/node_modules/gts/build/src/cli.js:33:23)
    at Object.next (/Users/beckwith/.nvm/versions/node/v8.9.1/lib/node_modules/gts/build/src/cli.js:14:53)
    at /Users/beckwith/.nvm/versions/node/v8.9.1/lib/node_modules/gts/build/src/cli.js:8:71
    at new Promise (<anonymous>)
    at __awaiter (/Users/beckwith/.nvm/versions/node/v8.9.1/lib/node_modules/gts/build/src/cli.js:4:12)
    at run (/Users/beckwith/.nvm/versions/node/v8.9.1/lib/node_modules/gts/build/src/cli.js:52:12)
```

jinwoo:
Does it happen only when you explicitly pass the files?

JustinBeckwith:
It seems to only happen when I pass arguments in the folder/*.ext format.  Running `gts check lib/discovery.ts` works fine.  

Same bug applies to `fix`.

jinwoo:
Are you on windows?

JustinBeckwith:
lol no

jinwoo:
This sounds like the wildcard expansion doesn't happen from the shell. Let me take a look.

jinwoo:
[...]

PR #87, opened 2017-11-14
fix: ignore files that are not found

jinwoo:
This is needed when files are explicitly passed to
gts. Just ignore them.

This addresses https://github.com/google/ts-style/issues/86


jinwoo:
Added a test case.

jinwoo:
Can someone merge this? I don't seem authorized for that.

Issue #83, opened 2017-11-13
`gts check` doesn't work when installed globally


JustinBeckwith:
I am trying to use gts by doing an `npm install -g gts`.  I got this library that's going to be a train wreck to upgrade, so I want to do it one piece at a time with a global tool.  So I `cd` into my app dir, and run the check command, and this is the call stack:

```
beckwith-macbookpro:google-api-nodejs-client beckwith$ gts check
FatalError: Could not find config file at: /Users/beckwith/Code/google-api-nodejs-client/node_modules/gts/tslint.json
    at new FatalError (/Users/beckwith/.nvm/versions/node/v8.9.1/lib/node_modules/gts/node_modules/tslint/lib/error.js:27:28)
    at findConfigurationPath (/Users/beckwith/.nvm/versions/node/v8.9.1/lib/node_modules/gts/node_modules/tslint/lib/configuration.js:55:19)
    at Object.findConfiguration (/Users/beckwith/.nvm/versions/node/v8.9.1/lib/node_modules/gts/node_modules/tslint/lib/configuration.js:41:22)
    at lint (/Users/beckwith/.nvm/versions/node/v8.9.1/lib/node_modules/gts/build/src/lint.js:31:48)
    at /Users/beckwith/.nvm/versions/node/v8.9.1/lib/node_modules/gts/build/src/cli.js:77:46
    at step (/Users/beckwith/.nvm/versions/node/v8.9.1/lib/node_modules/gts/build/src/cli.js:33:23)
    at Object.next (/Users/beckwith/.nvm/versions/node/v8.9.1/lib/node_modules/gts/build/src/cli.js:14:53)
    at /Users/beckwith/.nvm/versions/node/v8.9.1/lib/node_modules/gts/build/src/cli.js:8:71
    at new Promise (<anonymous>)
    at __awaiter (/Users/beckwith/.nvm/versions/node/v8.9.1/lib/node_modules/gts/build/src/cli.js:4:12)
```

PR #88, opened 2017-11-14
fix: correct gts root directory

ofrobots:
Fixes: https://github.com/google/ts-style/issues/83

Done ~~We need tests for use as a global install module.~~

Done #89 ~~It is worth noting that we do not expect `gts compile` to work from a globally installed version. We need to detect that case and show an error to the user in that scenario.~~

### DIFF
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -14,6 +14,7 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
+import * as path from 'path';
 import * as meow from 'meow';
 import * as updateNotifier from 'update-notifier';
 import {init} from './init';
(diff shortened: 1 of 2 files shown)


ofrobots:
Added test. This is ready for review.

src/cli.ts
```diff
@@ -71,7 +72,7 @@ function usage(msg?: string): void {
 async function run(verb: string, files: string[]): Promise<boolean> {
   const options: Options = {
     dryRun: cli.flags.dryRun || false,
-    gtsRootDir: `${process.cwd()}/node_modules/gts`,
+    gtsRootDir: path.resolve(__dirname, '../..'),
```
jinwoo: I was having a hard time understanding this. Worth commenting that `cli.js` is installed under `{root}/build/src/`?

ofrobots: This is valid criticism. However, I think this happens pervasively across the codebase, include when the `package.json`, `tslint.json` and other files are loaded. Adding a comment here but not elsewhere doesn't seem right, and neither does it make sense to have these littered throughout the code-base.

Perhaps we should switch the directory structure that doesn't have this problem (i.e. having a toplevel `ts/` directory that would contain `src` and `test`). WDYT?

jinwoo: Yeah, understood. Having `src` and `test` under `ts` would help but I don't think it's a high priority.

Issue #90, opened 2017-11-20
Running `npm install` after `npx gts init`


JustinBeckwith:
So I've done this a few times now.  I will run `npx gts init`, then immediately try doing a `npm run compile`.  This always fails with an error like this:

```
error TS5058: The specified path does not exist:
 '/users/beckwith/code/testy/node_modules/gts/tsconfig-google.json'.
```

Then I remember to run `npm install`, and everything is fine.  Is it possible to kick off an `npm install` after `npx gts init` modifies my `package.json`?

ofrobots:
+1. Good idea. PRs welcome :).

PR #92, opened 2017-11-22
feat: run npm install after init

JustinBeckwith:
Addresses #90

### DIFF
--- a/src/init.ts
+++ b/src/init.ts
@@ -38,6 +38,8 @@ async function query(
     options: Options): Promise<boolean> {
   if (options.yes) {
     return true;
+  } else if (options.no) {
+    return false;
   }
 
   if (message) {
@@ -215,5 +217,11 @@ export async function init(options: Options): Promise<boolean> {
     options.logger.log('No edits needed in package.json.');
   }
   await generateTsConfig(options);
+
+  // Run `npm install` after initial setup so `npm run check` works right away.
+  if (!options.dryRun) {
+    cp.spawnSync('npm', ['install'], {stdio: 'inherit'});
+  }
+
   return true;
 }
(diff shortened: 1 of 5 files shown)


src/init.ts
```diff
@@ -215,5 +215,9 @@ export async function init(options: Options): Promise<boolean> {
     options.logger.log('No edits needed in package.json.');
   }
   await generateTsConfig(options);
+
+  // Run `npm install` after initial setup so `npm run check` works right away
+  cp.spawnSync('npm', ['install'], {stdio: 'inherit'});
```
ofrobots: Question:
* Should this happen if we didn't actually modify dependencies? My guess is yes.
* Should this happen if the user passed `--dry-run`? My guess is no.
* Can you add a test to `test-kitchen`?

JustinBeckwith: Answers!
- Yes, I think so.
- Nope, fixed :)
- Done.  Was more of an adventure than I expected.

JustinBeckwith:
@ofrobots testing this turned into an adventure.  Turns out we weren't really testing `npx init` at all.  To make that work, I needed to add a `-n` flag.  PTAL!

test/test-kitchen.ts
```diff
@@ -71,11 +71,18 @@ test.before(async () => {
...
   await ncpp('test/fixtures', `${stagingPath}/`);
-  await simpleExecp('npm install', execOpts);
 });
 
 test.serial('init', async t => {
-  await simpleExecp('./node_modules/.bin/gts init -y', execOpts);
+  const nodeVersion = Number(process.version.slice(1).split('.')[0]);
+  if (nodeVersion < 8) {
+    await simpleExecp('npm install', execOpts);
+    await simpleExecp('./node_modules/.bin/gts init -y', execOpts);
```
ofrobots: Q: Any reason why there is `-y` on this leg of the branch and a `-n` on the other? If there is a valid reason, this deserves a comment.

JustinBeckwith: Added a comment.  You could use -y or -n in the < 8 version, but you must use -n in the > 8 version.

PR #97, opened 2017-12-05
feat: show clang-format version in gts check

kjin:


test/test-kitchen.ts
```diff
@@ -114,7 +114,7 @@ test.serial('generated json files should terminate with newline', async t => {
 test.serial('check before fix', async t => {
   const {exitCode, stdout} = await execp('npm run check', execOpts);
   t.deepEqual(exitCode, 1);
-  t.notDeepEqual(stdout.indexOf('clang-format reported errors'), -1);
+  t.regex(stdout, /clang-format \d+.\d+.\d+ reported errors/);
```
jinwoo: I think this should be `\d+\.\d+\.\d+`. I.e. a backslash before each `.`

JustinBeckwith:
I'm not opposed to landing this, but I do want to have a discussion.  It's easy for things like debugging information and versions like this to slowly build up, and run amok (see `npm`).  Instead of adding this here, what do folks think of having a `--verbose` flag or something that emits more of this kind of information?

src/format.ts
```diff
@@ -41,8 +42,8 @@ export async function format(
   } else {
     const result = await checkFormat(srcFiles);
     if (!result) {
-      options.logger.log(
-          'clang-format reported errors... run `gts fix` to address.');
+      options.logger.log(`clang-format ${
+          clangFormatVersion} reported errors... run \`gts fix\` to address.`);
```
ofrobots: When is this version information useful?

kjin: `clang-format` doesn't follow semver. If we fix the version number this isn't an issue, but it is still good to know. Even when the issue isn't clang-format, it is nice to be able to tell that at a glance.

kjin:
@JustinBeckwith That sounds good to me.

kjin:
Closing this in favor of #100 (fix clang-format version).

PR #99, opened 2017-12-06
fix: format root files only

### LINKED ISSUE #98: `gts check/fix` uses a list of files created by tslint
See below comment

~On my mac, `gts check` exits with code 0 before checking all files [here](https://github.com/GoogleCloudPlatform/cloud-trace-nodejs/pull/621). On linux this doesn't happen, and I get my clang-format error reports.~

~I combed `gts` dependencies for `process.exit` calls but couldn't find any. I believe unresolved Promises are somehow being ignored, and the event loops drains fully. Unsure about this...~

kjin:
Fixes #98

We should use `getRootFileNames` instead of `getSourceFiles` when retrieving a list of files to format; the latter gets every files in the dependency graph constructed from files included in `tsconfig.json`.

### DIFF
--- a/src/format.ts
+++ b/src/format.ts
@@ -31,13 +31,10 @@ export async function format(
     options: Options, files: string[] = [], fix = false): Promise<boolean> {
   const program = createProgram(options);
   const srcFiles = files.length > 0 ?
-      files :
-      program.getSourceFiles()
-          .map(sourceFile => sourceFile.fileName)
-          .filter(f => !f.endsWith('.d.ts'));
+      files : program.getRootFileNames().filter(f => !f.endsWith('.d.ts'));
 
   if (fix) {
-    return fixFormat(srcFiles);
+    return await fixFormat(srcFiles);
   } else {
     const result = await checkFormat(srcFiles);
     if (!result) {
--- a/src/format.ts
+++ b/src/format.ts
@@ -30,6 +30,12 @@ const baseArgs =
 export async function format(
     options: Options, files: string[] = [], fix = false): Promise<boolean> {
   const program = createProgram(options);
+  // Obtain a list of source files to format.
+  // We use program.getRootFileNames to get only the files that match the
+  // include patterns specified in the given tsconfig.json file (as specified
+  // through options). This is necessary because we only want to format files
+  // over which the developer has control (i.e. not auto-generated or
+  // third-party source files).
   const srcFiles = files.length > 0 ?
       files :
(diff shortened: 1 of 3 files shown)


src/format.ts
```diff
@@ -73,6 +71,7 @@ function fixFormat(srcFiles: string[]): Promise<boolean> {
  * @param srcFiles list of source files
  */
 function checkFormat(srcFiles: string[]): Promise<boolean> {
+  console.log(srcFiles);
```
ofrobots: Errant log?

test/fixtures/kitchen/src/server.ts
```diff
@@ -1 +1 @@
-  const isThisTypeScript = true
+  export const isThisTypeScript = true
```
ofrobots: why was this change needed?

test/test-format.ts
```diff
@@ -0,0 +1,59 @@
...
+  yes: false,
+  no: false,
+  logger: {log: console.log, error: console.error, dir: nop}
+};
+
+test.serial('format should return false for ill-formatted files', async t => {
+  await withFixtures(
+      {
+        'tsconfig.json': JSON.stringify({files: ['a.ts']}),
+        'a.ts': 'export const foo = 2;  '
```
jinwoo: Is this ill-formatted because of the trailing spaces? Can you add a comment for that so that future readers don't accidentally delete the spaces?

jinwoo: Actually I should've put this comment in line 51 below because if someone accidentally removes the trailing spaces there, the test will still pass.

kjin:
@jinwoo @ofrobots PTAL

src/format.ts
```diff
@@ -32,12 +32,10 @@ export async function format(
   const program = createProgram(options);
   const srcFiles = files.length > 0 ?
       files :
-      program.getSourceFiles()
-          .map(sourceFile => sourceFile.fileName)
-          .filter(f => !f.endsWith('.d.ts'));
+      program.getRootFileNames().filter(f => !f.endsWith('.d.ts'));
```
jinwoo: Can you briefly comment here on why only root files should be formatted?

test/test-format.ts
jinwoo: I'd put the same comment here too so that people can easily understand why this line must fail :)

src/format.ts
kjin: Done

PR #100, opened 2017-12-11
chore: upgrade and pin clang-format at 1.2.1

kjin:
Supersedes #97

The `clang-format` module incorporates changes to formatting rules as semver patch-level changes. This is undesirable for users of our module, as we treat changes in formatting as semver-major. This PR addresses this by pinning `clang-format` to 1.2.1; going forward we should increment the pinned version with each semver-major bump of this module.

### DIFF
--- a/package.json
+++ b/package.json
@@ -39,7 +39,7 @@
   "license": "Apache-2.0",
   "dependencies": {
     "chalk": "^2.3.0",
-    "clang-format": "^1.0.53",
+    "clang-format": "1.2.1",
     "inquirer": "^3.2.1",
     "meow": "^4.0.0",
     "pify": "^3.0.0",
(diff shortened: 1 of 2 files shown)


ofrobots:
The PR and commit message should document the rationale for why the semver range is being locked down.

package.json
```diff
@@ -39,7 +39,7 @@
   "license": "Apache-2.0",
   "dependencies": {
     "chalk": "^2.3.0",
-    "clang-format": "^1.0.53",
+    "clang-format": "1.2.1",
```
JustinBeckwith: Why no `^`?

kjin: `clang-format` doesn't follow semver. 1.0.53 and 1.0.55 for example don't output the same string. By fixing it we can prevent inconsistencies between installations made at different times or different versions of node.

PR #101, opened 2017-12-13
fix(init): use ~ semver range for typescript

ofrobots:
TypeScript doesn't follow semver, but we do. It makes sense to default
to a ~ semver range for typescript in the client projects, rather them
letting them stumble into the subtlety.

### DIFF
--- a/src/init.ts
+++ b/src/init.ts
@@ -93,7 +93,7 @@ async function addScripts(
 async function addDependencies(
     packageJson: PackageJson, options: Options): Promise<boolean> {
   let edits = false;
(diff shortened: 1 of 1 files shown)


src/init.ts
```diff
@@ -93,7 +93,7 @@ async function addScripts(
 async function addDependencies(
     packageJson: PackageJson, options: Options): Promise<boolean> {
   let edits = false;
-  const deps: Bag<string> = {'gts': `^${pkg.version}`, 'typescript': '^2.6.1'};
+  const deps: Bag<string> = {'gts': `^${pkg.version}`, 'typescript': '~2.6.1'};
```
jinwoo: Doesn't matter much but 2.6.2 is the latest version.

PR #102, opened 2017-12-13
build: switch to circleci

ofrobots:


ofrobots:
This PR is ready for review.

JustinBeckwith:
Just curious, why?

ofrobots:
1. faster
1. publish to npm workflow with travis seems buggy.
1. right at github you can see whether the tests are failing against all versions of node, or only certain.

Nothing earth shattering, but we want to setup automatic publishing of the npm modules when a tag is pushed. It would be good have use the same config/setup for this across all the repos we manage, and we already have some setup for this for CircleCI.

PR #104, opened 2018-01-02
fix(package): update inquirer to version 4.0.2

ofrobots:
Closes #96


ofrobots:
Abandoning this PR. `inquirer@4` has dropped support for Node 4. We can upgrade when we drop support for Node 4 ourselves – around April 2018.

PR #105, opened 2018-01-02
chore: add license checking with js-green-license

jinwoo:
Run `jsgl` on the repository as one of posttest checks.


jinwoo:
I realize I have to `npm publish js-green-licenses` again to support Node 6 here. Not sure how not to run `jsgl` for Node 4 though.

jinwoo:
Now using `js-green-licenses` 0.1.2. It supports Node 4 & 6.

Issue #111, opened 2018-01-10
Test 'kitchen › use as a non-locally installed module' fails


jinwoo:
I'm not sure since when this test started to fail.

jinwoo:
Fixed with #112

PR #112, opened 2018-01-11
fix: fix a failing test

jinwoo:
`gts init` must be done before `gts check`.

Not sure how this has been passing so far. This started to fail
suddenly.


jinwoo:
This fixes https://github.com/google/ts-style/issues/111.

Issue #113, opened 2018-01-11
`gts init` uses globally installed gts path for generated tsconfig.json


jinwoo:
With the latest version `gts`, if I use a globally installed one to generate a new project with `gts init`, the generated `tsconfig.json` looks like this.

```json
{
  "extends": "./../.nvm/versions/node/v9.3.0/lib/node_modules/gts/tsconfig-google.json",
  "compilerOptions": {
    "rootDir": ".",
    "outDir": "build"
  },
  "include": [
    "src/*.ts",
    "src/**/*.ts",
    "test/*.ts",
    "test/**/*.ts"
  ],
  "exclude": [
    "node_modules"
  ]
}
```

See the first line. It should use a locally installed gts path:
```json
{
  "extends": "./node_modules/gts/tsconfig-google.json",
  ...
}
```

jinwoo:
Fixed with #114

PR #114, opened 2018-01-11
fix: use local gts path for generated tsconfig.json

jinwoo:
Currently the global gts path is used when `gts init` is done using a
globally installed `gts`, which is a usual workflow.

### DIFF
--- a/src/init.ts
+++ b/src/init.ts
@@ -152,13 +152,8 @@ async function generateTsConfig(options: Options): Promise<void> {
     }
   }
 
-  // typescript expects relative paths to start with './'.
-  const baseConfig = './' +
-      path.relative(
-          options.targetRootDir,
-          path.resolve(options.gtsRootDir, 'tsconfig-google.json'));
   const tsconfig = formatJson({
-    extends: baseConfig,
+    extends: './node_modules/gts/tsconfig-google.json',
     compilerOptions: {rootDir: '.', outDir: 'build'},
     include: ['src/*.ts', 'src/**/*.ts', 'test/*.ts', 'test/**/*.ts'],
     exclude: ['node_modules']
--- a/test/test-kitchen.ts
+++ b/test/test-kitchen.ts
@@ -85,8 +85,13 @@ test.serial('init', async t => {
         `npx -p ${stagingPath}/gts.tgz --ignore-existing gts init -n`,
         execOpts);
   }
-  fs.accessSync(`${stagingPath}/kitchen/tsconfig.json`);
-  t.pass();
+  const tsconfigPath = `${stagingPath}/kitchen/tsconfig.json`;
+  fs.accessSync(tsconfigPath);
+
+  // The `extends` field must use the local gts path.
+  const tsconfigJson = fs.readFileSync(tsconfigPath, 'utf8');
+  const tsconfig = JSON.parse(tsconfigJson);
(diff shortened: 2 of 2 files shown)


jinwoo:
One test will fail, which is being fixed by https://github.com/google/ts-style/pull/112

jinwoo:
This fixes https://github.com/google/ts-style/issues/113.

test/test-kitchen.ts
```diff
@@ -95,10 +95,15 @@ test.serial('use as a non-locally installed module', async t => {
   const GTS = `${stagingPath}/kitchen/node_modules/.bin/gts`;
   const tmpDir = tmp.dirSync({keep, unsafeCleanup: true});
   await ncpp('test/fixtures', `${tmpDir.name}/`);
+
   const opts = {cwd: `${tmpDir.name}/kitchen`};
-  // It's important to use `-n` here because we don't want to overwrite
-  // the version of gts installed, as it will trigger the npm install.
-  await simpleExecp(`${GTS} init -n`, opts);
```
jinwoo: I changed `-n` to `-y` because a local `gts` must be installed so that the `gts check` below can find the tsconfig-google.json file.

jinwoo:
As we discussed, I moved the test to 'use as a non-locally installed module'.

src/init.ts
```diff
@@ -152,13 +152,8 @@ async function generateTsConfig(options: Options): Promise<void> {
...
   }
 
-  // typescript expects relative paths to start with './'.
-  const baseConfig = './' +
-      path.relative(
-          options.targetRootDir,
-          path.resolve(options.gtsRootDir, 'tsconfig-google.json'));
   const tsconfig = formatJson({
-    extends: baseConfig,
+    extends: './node_modules/gts/tsconfig-google.json',
```
DominicKramer: Should `path.join` instead of using `/` directly?

test/test-kitchen.ts
```diff
@@ -95,10 +95,15 @@ test.serial('use as a non-locally installed module', async t => {
...
+  await simpleExecp(`${GTS} init -y`, opts);
+  // The `extends` field must use the local gts path.
+  const tsconfigJson =
```
DominicKramer: nit: Could you use `pify` to promisify `readFile` so that you can `await` the `readFile`?  Also, should you be using `path.join` here and in the `t.deepEqual` line?

src/init.ts
jinwoo: I don't think we should have a different value here depending on on which platform the source code is generated, right?

test/test-kitchen.ts
jinwoo: This test file uses lots of `...Sync` functions. I agree that using async functions is better but since this is a test file, I think we'd better make the code more consistent. WDYT?

DominicKramer: I agree its not that big of a deal since its a test file, especially if there are a lot of places where `*Sync` functions are used.  If we wanted to make them async, it would probably be best as a follow-up PR.

src/init.ts
DominicKramer: I'm not sure.  I would expect `tsconfig` to be like `package.json` where we can always use `/`, but we should probably make sure.

jinwoo: Windows has always supported `/` as a path separator as well as `\`. So I don't see any issues with this if we consider all the current major OS's.

DominicKramer: Ok sounds good.

PR #124, opened 2018-02-05
feat: use typescript 2.7.1 and esModuleInterop

jinwoo:
This PR is both for the generated `package.json` and `gts` itself.

TypeScript 2.7.1 added `--esModuleInterop` for better ES module
support. And for that, we need to turn on `allowSyntheticDefaultImports`
too.

Also `tslint` must be upgraded to 5.9.1 for it to compile with
`esModuleInterop`.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -80,6 +80,64 @@
         "arrify": "1.0.1"
       }
     },
+    "@ladjs/time-require": {
+      "version": "0.1.4",
+      "resolved": "https://registry.npmjs.org/@ladjs/time-require/-/time-require-0.1.4.tgz",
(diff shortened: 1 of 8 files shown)


ofrobots:
Typescript 2.7.x would be a semver major change; so let's no land this now. It would be okay landing this onto a `next` branch though.

jinwoo:
@ofrobots Yes, you're right that this is a breaking change. What about releasing 0.6.0 after this change? I'm fine with landing onto `next` if it's not a good idea.

BTW I think we need a release for #114.

jinwoo:
Closing this. New PR in #125.

PR #125, opened 2018-02-06
feat: use typescript 2.7.1 and esModuleInterop

jinwoo:
[BREAKING CHANGE]

This PR is both for the generated `package.json` and `gts` itself.

TypeScript 2.7.1 added `--esModuleInterop` for better ES module
support. And for that, we need to turn on `allowSyntheticDefaultImports`
too.

Also `tslint` must be upgraded to 5.9.1 for it to compile with
`esModuleInterop`.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -80,6 +80,64 @@
         "arrify": "1.0.1"
       }
     },
+    "@ladjs/time-require": {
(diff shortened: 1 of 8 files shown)


jinwoo:
This replaces #124. This PR will land the change onto the `next` branch.

PR #126, opened 2018-02-28
cross platform friendly

ndxbn:
I cannot execute `npm run clean` on Windows.

This  pull request makes npm-script to cross platform friendly.

### DIFF
--- a/package.json
+++ b/package.json
@@ -20,7 +20,7 @@
   ],
   "scripts": {
     "build": "npm run compile",
-    "clean": "shx rm -rf ./build/",
+    "clean": "rimraf ./build/",
     "codecov": "nyc report --reporter=json && codecov -f coverage/*.json",
     "compile": "tsc -p .",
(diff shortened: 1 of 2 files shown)


ndxbn:
I signed it!

package.json
```diff
@@ -20,7 +20,7 @@
   ],
   "scripts": {
     "build": "npm run compile",
-    "clean": "rm -rf ./build/",
+    "clean": "shx rm -rf ./build/",
```
jinwoo: Thank you for the contribution!

`shx` seems like a good utility but we already have `rimraf` installed for doing `rm -rf`. Can you please just use it instead of installing a new package? Thanks!

ndxbn: ok.

well, should I rebase? ( https://github.com/google/ts-style/pull/126/commits/ff140de4fc550ac88daa1cfb5b5c5bc398ea0c85 commit goese away )

Issue #128, opened 2018-03-02
Error in tsconfig.json after running `npx gts init`


doplumi:
`The specified path does not exist: '/users/dom/.npm/_npx/98679/lib/node_modules/gts/tsconfig-google.json'.`

jinwoo:
I believe this was fixed by https://github.com/google/ts-style/pull/114. But we haven't released a new version yet.

@ofrobots Can we release a new version?

doplumi:
@jinwoo great, plus I just noticed, it downloads Chromium for some reason? I've got puppeteer as a dependency, but what does gts wants to do with it?

doplumi:
Also – going back on topic – it doesn't seem to install any `tslint.json`. Is that part of the same issue @jinwoo?

jinwoo:
`gts` doesn't download Chromium by itself. What happens is `gts init` runs `npm install` after initializing the project and because your projects depends on puppeteer, `npm install` downloads Chromium.

jinwoo:
`gts init` doesn't install `tslint.json` in your project's directory. It just uses the `tslint.json` file in the `gts` directory.

jinwoo:
0.5.4 has been published. Please check with the new version. Thanks again for the bug report. Closing now.

PR #129, opened 2018-03-02
chore: release 0.5.4

jinwoo:
This release includes a fix for
https://github.com/google/ts-style/issues/128.

The draft release notes are in https://github.com/google/ts-style/releases/tag/untagged-8ad589cd37ce5d87c562

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "gts",
-  "version": "0.5.3",
+  "version": "0.5.4",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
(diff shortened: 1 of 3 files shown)


jinwoo:
One test is failing. I'm not sure why that test is suddenly failing.

jinwoo:
I fixed the failing test. That test has been always tricky for me. It was really hard to understand why the test suddenly fails. Now I understand the test better and know why it started to fail.

The test did `gts init -y`, which overwrites the `package.json` file and because I'm bumping up the `gts` version, `npm install` fails to find that version to install. We need to install `gts` from the local tarball instead.

I think now everything is ok with that test.

PR #140, opened 2018-03-29
feat: upgrade to Typescript 2.8.x

ofrobots:
This is for the next branch.

Update the typescript dependency to 2.8.x both for the locally used dev
dependency and the version we emit in the generated package.json. The
peerDependency remains unchanged, which makes this a semver-minor.

Existing projects already have a dependency on typescript in their
package.json, so they should continue using the pre-2.8 version.

New versions projects will get 2.8.1+.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -136,6 +136,16 @@
         "@types/node": "9.3.0"
       }
     },
+    "@types/inquirer": {
+      "version": "0.0.40",
+      "resolved": "https://registry.npmjs.org/@types/inquirer/-/inquirer-0.0.40.tgz",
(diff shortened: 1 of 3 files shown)


jinwoo:
The `next` branch is already semver-major due to esModuleInterop. Why do we have to be semver-minor?

ofrobots:
Each change can have its own semver impact. This one has a semver-minor impact. Since the previous change is semver major, it will just get absorbed into the bump for that one.

Independently, I presently don't have a reason to make this change a major by requiring that subject projects have a peer dependency on 2.8.x. There are no compile time flags that we have changed in `tsconfig.json` that force users of this module to use 2.8 or newer. 2.7 is just fine.

PR #142, opened 2018-03-29
fix: don't compile right after generating a project

jinwoo:
Compilation would fail because there's no source files.

### DIFF
--- a/src/init.ts
+++ b/src/init.ts
@@ -215,7 +215,9 @@ export async function init(options: Options): Promise<boolean> {
 
   // Run `npm install` after initial setup so `npm run check` works right away.
   if (!options.dryRun) {
-    cp.spawnSync('npm', ['install'], {stdio: 'inherit'});
+    // --ignore-scripts so that compilation doesn't happen because there's no
+    // source files yet.
+    cp.spawnSync('npm', ['install', '--ignore-scripts'], {stdio: 'inherit'});
   }
 
   return true;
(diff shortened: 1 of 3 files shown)


src/init.ts
```diff
@@ -215,7 +215,9 @@ export async function init(options: Options): Promise<boolean> {
 
   // Run `npm install` after initial setup so `npm run check` works right away.
   if (!options.dryRun) {
-    cp.spawnSync('npm', ['install'], {stdio: 'inherit'});
+    // --ignore-scripts so that compilation doesn't happen because there's no
+    // source files yet.
+    cp.spawnSync('npm', ['install', '--ignore-scripts'], {stdio: 'inherit'});
```
DominicKramer: If someone has a `postinstall` script, I don't think that will get invoked either.  However, I expect this use case is probably not too common and, if needed, someone can always run `npm install` manually if needed.

test/test-kitchen.ts
```diff
@@ -113,6 +113,10 @@ test.serial('use as a non-locally installed module', async t => {
   // server.ts has a lint error. Should error.
   await t.throws(simpleExecp(`${GTS} check src/server.ts`, opts));
 
+  // Compilation shouldn't have happened. Hence no `build` directory.
+  const dirContents = fs.readdirSync(`${tmpDir.name}/kitchen`);
+  t.is(dirContents.indexOf('build'), -1);
```
DominicKramer: Is it possible that `npm run compile` was run but had an error that caused the `build` directory to not be created?

src/init.ts
jinwoo: Yeah, most of the cases `gts init` is run from an empty directory. Even though someone had a `postinstall` script, they would know what they are doing and be able to figure things out.

test/test-kitchen.ts
jinwoo: I verified that the `build` directory is created without my change.

jinwoo:
I moved the new test to a better place.

Issue #70, opened 2017-10-27
Why does the generated tsconfig.json contain both an includes and excludes section?


ofrobots:
For context: https://github.com/google/google-p12-pem/pull/9#discussion_r147466220

```
+  "include": [
 +    "src/*.ts",
 +    "src/**/*.ts",
 +    "test/*.ts",
 +    "test/**/*.ts"
 +  ],
+  "exclude": [
 +    "node_modules"
 +  ]
```

> I think you can leave 'include' out and it just compiles **/*.ts (obeying the exclude).
> Alternatively, you can leave the exclude out. Specifying both is redundant I think.

IIRC, `gts` sets things up this way because of some trouble we some trouble with `tsc` ending up looking into `node_modules` despite the `includes` section. I can't remember the details at the moment however. We should revisit and document the reason for this.

PR #143, opened 2018-04-03
fix: exclude section is not necessary

ofrobots:
tsconfig.json doesn't need an exclude section when an include section
is present.

Fixes: https://github.com/google/ts-style/issues/70

### DIFF
--- a/src/init.ts
+++ b/src/init.ts
@@ -155,8 +155,7 @@ async function generateTsConfig(options: Options): Promise<void> {
   const tsconfig = formatJson({
     extends: './node_modules/gts/tsconfig-google.json',
     compilerOptions: {rootDir: '.', outDir: 'build'},
-    include: ['src/*.ts', 'src/**/*.ts', 'test/*.ts', 'test/**/*.ts'],
-    exclude: ['node_modules']
+    include: ['src/*.ts', 'src/**/*.ts', 'test/*.ts', 'test/**/*.ts']
   });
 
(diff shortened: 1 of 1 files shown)


src/init.ts
```diff
@@ -155,8 +155,7 @@ async function generateTsConfig(options: Options): Promise<void> {
   const tsconfig = formatJson({
     extends: './node_modules/gts/tsconfig-google.json',
     compilerOptions: {rootDir: '.', outDir: 'build'},
-    include: ['src/*.ts', 'src/**/*.ts', 'test/*.ts', 'test/**/*.ts'],
-    exclude: ['node_modules']
```
jinwoo: I had to exclude `node_modules` in one of my projects sometime in the past, but I forgot why.

src/init.ts
```diff
@@ -155,8 +155,7 @@ async function generateTsConfig(options: Options): Promise<void> {
...
+    include: ['src/*.ts', 'src/**/*.ts', 'test/*.ts', 'test/**/*.ts']
```
jinwoo: Side note. We say `main: index.js` when we say our source files are in these directories... well...

ofrobots: That's because we generate the `package.json` using `npm` defaults hence:

https://github.com/ofrobots/ts-style/blob/4290d772df259e886728770f3cb77fba8eb8a2db/src/init.ts#L199

We should refine this to better match the expected project layout.

PR #146, opened 2018-04-23
chore: drop support for Node 4

ofrobots:
BREAKING CHANGE. Stop supporting Node 4.x.

### DIFF
--- a/package.json
+++ b/package.json
@@ -33,7 +33,7 @@
     "license-check": "jsgl --local ."
   },
   "engines": {
-    "node": ">=4"
+    "node": ">6"
   },
   "keywords": [],
   "author": "Google Inc.",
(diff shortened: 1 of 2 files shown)


JustinBeckwith:
May as well bump `target` in `tsconfig.js` to at least `es2016`.  It's even possible `es2017` works here 🥁

ofrobots:
I think that can be discussed / addressed in a PR on its own. Let's keep this one focussed on the bare minimum for dropping node 4 support.

ofrobots:
Extracted target setting discussion into https://github.com/google/ts-style/issue/147.

package.json
```diff
@@ -33,7 +33,7 @@
     "license-check": "jsgl --local ."
   },
   "engines": {
-    "node": ">=4"
+    "node": ">6"
```
jinwoo: Does `>6` include 6?

ofrobots: ugh.. not sure how I did that typo

PR #155, opened 2018-05-08
chore: release 0.6.0

JustinBeckwith:
[Release notes](https://github.com/google/ts-style/releases/edit/untagged-34e192980be6eb650e6b)

As it would turn out, clang-format 1.2.2 has a nasty TypeScript formatting bug, and the fix is in 1.2.3.  It's probably time to ship :)

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "gts",
-  "version": "0.5.4",
+  "version": "0.6.0",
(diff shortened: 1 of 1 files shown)


jinwoo:
Err... https://github.com/google/ts-style/pull/125 was a breaking change and will break some builds, I think.

PR #157, opened 2018-05-19
chore: start testing against Node 10

ofrobots:


.circleci/config.yml
```diff
@@ -50,15 +50,15 @@ jobs:
...
     docker:
-      - image: node:9
+      - image: node:10
         user: node
     <<: *unit_tests
 
   publish_npm:
     docker:
-      - image: node:8
+      - image: node:10
```
jinwoo: It wouldn't matter much but isn't it still better to publish with 8 because it's LTS?

ofrobots: Modules published with 10, should work with other versions of Node too, so is there a tangible difference here?

jinwoo: Not really. LGTM.

PR #158, opened 2018-06-01
test: resolve gtsRootDir in tests

ofrobots:
This ensures that the tests are independent of where they are run from.
Remove a test that was unnecessary.

### DIFF
--- a/test/test-lint.ts
+++ b/test/test-lint.ts
@@ -15,6 +15,7 @@
  */
 
 import test from 'ava';
+import * as path from 'path';
 
 import {Options} from '../src/cli';
 import * as lint from '../src/lint';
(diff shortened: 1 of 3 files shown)


test/test-lint.ts
```diff
@@ -34,36 +35,18 @@ const OPTIONS: Options = {
...
-const TSLINT_CONFIG = require('../../tslint.json');
-
 test.serial('createProgram should return an object', async t => {
   await withFixtures({'tsconfig.json': '{}'}, async () => {
     const program = lint.createProgram(OPTIONS);
     t.truthy(program);
   });
 });
 
-// TODO: make the error friendlier. This error implies that our module messed
```
jinwoo: Why is this test deletec?

ofrobots: The tests in this fine weren't fully thought through, and this one in particular wasn't needed. Instead of fixing the bug in `lint.ts`, the tests were always ensuring that there was a `tslint.json` in the specific test directory. This test was testing that the bug correctly existed 🤣.

ofrobots: Sorry, this should had said.. fix the bug in the `OPTIONS` object construction.

PR #159, opened 2018-06-01
feat: ability to use user provided tslint config

ofrobots:
This change adds the ability to use tslint.conf provided by the user.
If the user doesn't have a config file, we fall back to the default
gts config file.

This depends on #158, so ignore that commit when reviewing. That commit will be removed once #158 lands.

We do prefer the user-provided config file over the default now, which is potentially breaking, but I doubt users ever had the reason to have a local `tslint.json` before. I don't think this is semver major, but feel free to push back.


ofrobots:
@JustinBeckwith added some wording in the README. PTAL.

JustinBeckwith:
LGTM!

Issue #127, opened 2018-02-28
Add the ability to use arbitrary configuration files (TS and TSLint)


vyshkant:
I set up my project so that my TypeScript configuration file is different from `tsconfig.json`. In addition, I have my own TSLint config file and I want to use it.

At the same time I want to use `gts`, but don't see the way to specify the paths to my TS and TSLint configs.

Can you please tell me, can I do it?

If this requires further work, I will be happy to write the PR (but I will need your advice).

jinwoo:
Thanks for the suggestion!

Re: using a custom tslint config file, I don't think we want to support that. That's because `gts` is an *opinionated* style tool and we want it to enforce Google's typescript style guideline.

I think it is a good feature to support a different typescript config file though. But it seems `tsconfig.json` is used only for the `gts clean` command. I may be wrong.

ofrobots:
Once you run `gts init`, we create a `tsconfig.json` in the package directory that inherits from the Google TS Config. If you want to override that, you can customize the `tsconfig.json` as you desire, but note that some commands (e.g. `gts clean`) depend on `tsconfig.json` having certain keys.

Our intent was to have a consistent look and feel for all projects using the style guide. However given that we do allow some overriding of `tsconfig.json`, I would be supportive of a mechanism to override `tslint.json` as well – as long as we don't encourage users to override.

If you're willing to put together a PR, we'll take a look!

TheGrimSilence:
Personally, I support the ability to override the lint configuration. I enjoy using gts, but there are a few things as a TypeScript React developer I simply need to override such as `semicolon` and `interface-name`. These are small things but quickly become frustrating when your entire workflow is set up for these things and then gts spits out linting errors and adds semicolons left and right.

Issue #119, opened 2018-01-25
Install a tslint.json that extends installed gts' tslint.json like the tsconfig.json does


usergenic:
Necessary for IDEs that have tslint integration.  Would also enable extensibility of the tslint config where needed while supporting keeping up-to-date with base config.

ofrobots:
We expose the `tsconfig.json` because it would have been impractical to assume that the needs of all projects can be satisfied with a canned `tsconfig.json`. For example, users typically need to edit the input files, output directories, etc.

The reason why we chose to not expose `tslint.json` is the thinking was that there ought to be very few reason to diverge from the style guide as far as the linting rules are concerned. So while exposing this possible, the intent was to discourage customization / extension.

The use case of IDEs is valid though.

hvent90:
@ofrobots, a desire to allow "any" lead me here, and I was wondering if you'd care to share your thoughts on this:

I want to be as strictly typed as possible. A hypothetical maintainer of an API may request to keep expectations open for how the response may alter or change. How could I avoid using an "any" type here?

I apologize if this is out of line for me to ask this here, but I would be extremely grateful to hear your thoughts!

keradus:
Then even more, let say I want to extend tslint by custom ruleset, eg Codelyzer, and use rules suggested by Angular himself, eg `no-conflicting-life-cycle-hooks` - without ability to adjust `tslint.json`, I can't do it with `ts-style`

PR #160, opened 2018-06-01
feat: generate tslint.json

ofrobots:
`init` now generates a tslint.json file which inherits from the default
configuration. This empowers the users to customize if necessary
instead of being overly opinionated.

Fixes: #127
Fixes: #119 

This depends upon #158 and #159, so the first two commits don't need to be reviewed. They will be removed once those PRs land.

### DIFF
--- a/README.md
+++ b/README.md
@@ -43,8 +43,6 @@ When you run the `npx gts init` command, it's going to do a few things for you:
   - `compile`: Compiles the source code using TypeScript compiler.
   - `pretest`, `posttest` and `prepare`: convenience integrations.
 
-We strongly recommend you use the default style config, but if you must tweak, you can edit the generated `tsconfig.json`. For linter, we use the default `tslint.json` unless we find that file in your project directory.
-
 ## Individual files
 The commands above will all run in the scope of the current folder.  Some commands can be run on individual files:
 
--- a/src/init.ts
+++ b/src/init.ts
@@ -142,40 +142,48 @@ async function writePackageJson(
   options.logger.dir(preview);
 }
 
-async function generateTsConfig(options: Options): Promise<void> {
-  let existing;
-  try {
-    existing = await read('./tsconfig.json', 'utf8');
-  } catch (err) {
-    if (err.code === 'ENOENT') {
-      /* not found, create it. */
-    } else {
-      throw new Error(`Unknown error reading tsconfig.json: ${err.message}`);
-    }
-  }
+async function generateTsLintConfig(options: Options): Promise<void> {
+  const config = formatJson({extends: 'gts/tslint.json'});
+  return generateConfigFile(options, './tslint.json', config);
+}
 
-  const tsconfig = formatJson({
+async function generateTsConfig(options: Options): Promise<void> {
+  const config = formatJson({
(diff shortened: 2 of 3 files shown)


JustinBeckwith:
I am 👎  on this approach.  My hope would be that the average user wouldn't need to have their own tslint.json.  I'm mostly 👎 on supporting the override at all, but I get why some people will just insist on it.  Instead of generating this during `init`, could we add a command that drops the file for you?  Like a `gts customize-lint` or something like that?

ofrobots:
This is a good discussion.

This is what the README says about this project:
> **No configuration**. The easiest way to enforce consistent style in your project. Just drop it in.

This also speaks to what you are trying to argue about the average user not needing to customize anything. I agree with that. Yet, this project, today, generates a `tsconfig.json` in the users's directory. Is that an invitation for the user to customize it? Maybe, maybe not (IMHO I don't think it is an invitation to customize). It is just how we wire things up. At the same time, we don't want this project to be too paternalistic and _prevent_ people from customizing `tsconfig.json`. By the same argument, I do not think that generating a `tslint.json` is an invitation for the user to customize the lint config.

Either way, we need to be consistent in our own reason. We need to have a good reason why it is okay to generate `tsconfig.json` and not the `tslint.json`. We should generate both, or neither.

JustinBeckwith:
Fair points!  I suppose in a perfect world, I'd prefer we take the same approach with `tsconfig.json`. As long as there was an easy way to generate it, not needing to have that either would be great :)

README.md
```diff
@@ -43,8 +43,6 @@ When you run the `npx gts init` command, it's going to do a few things for you:
   - `compile`: Compiles the source code using TypeScript compiler.
   - `pretest`, `posttest` and `prepare`: convenience integrations.
 
-We strongly recommend you use the default style config, but if you must tweak, you can edit the generated `tsconfig.json`. For linter, we use the default `tslint.json` unless we find that file in your project directory.
-
```
ofrobots: Intentionally removed. If we are going to generate `tslint.json` – we do not need to document further. Let's not invite customization.

ofrobots:
I think generating the files is friendlier. The value of this project is in providing a good, consistent style guide that can be used in a project without thinking about it.

If someone has thought about the precise, but different, style they want, and still find value in this project – let us not turn them away.

ofrobots:
> As long as there was an easy way to generate it, not needing to have that either would be great :)

I do not think this is practical. The user *needs* to be able to customize their repo structure.

ofrobots:
The other reason why this is desirable is mentioned in https://github.com/google/ts-style/issues/119: IDE support. A `tslint.json` in the correct location ends up getting picked up by the IDE or other tooling the user may have,.

I'm -1 on generating the configuration only under a flag, if provided by the user. That is not going to be discoverable. This needs to be the default behavior so that users don't have to pass a flag effectively equivalent to `--and-work-with-my-IDE`. I feel strong about this at this point.

JustinBeckwith:
Again, all fair points.  You're probably right about this, but I that doesn't mean I have to be happy about it 😆 You won me over.  I'm cool with this approach.

Issue #118, opened 2018-01-25
gts fix does not support --dry-run but CLI output makes it look like it should


usergenic:
If it can not be supported for this command, gts should throw an error if `gts fix --dry-run` is given. 
 If it can be supported, gts should output suggested changes.

PR #169, opened 2018-06-08
fix: fix with --dry-run should not modify files

ofrobots:
Fixes: https://github.com/google/ts-style/issues/118

### DIFF
--- a/test/test-lint.ts
+++ b/test/test-lint.ts
@@ -80,8 +80,25 @@ test.serial('lint should auto fix fixable errors', async t => {
       async (fixturesDir) => {
         const okay = lint.lint(OPTIONS, [], true);
         t.is(okay, true);
-        const contents = fs.readFileSync(path.join(fixturesDir, 'a.ts'), 'utf8');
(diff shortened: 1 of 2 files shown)


ofrobots:
Working on a few more tests

test/test-lint.ts
```diff
@@ -66,6 +71,37 @@ test.serial('lint should return false on bad code', async t => {
...
+      async (fixturesDir) => {
+        const okay = lint.lint(OPTIONS, [], true);
+        t.is(okay, true);
+        const contents =
+            fs.readFileSync(path.join(fixturesDir, 'a.ts'), 'utf8');
+        t.deepEqual(contents, FIXABLE_CODE_FIXED);
+      });
+});
+
+test.serial('lint should auto fix fixable errors unless dry-run', async t => {
```
jinwoo: The description sounds a little bit off from what is tested. Maybe 'lint should not auto fix on dry-run'?

ofrobots: done.

PR #176, opened 2018-06-18
fix(format): deal with absent .clang-format

ofrobots:
This fixes a regression in v0.7.0.

### DIFF
--- a/src/format.ts
+++ b/src/format.ts
@@ -13,12 +13,16 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
+import * as fs from 'fs';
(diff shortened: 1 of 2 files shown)


src/format.ts
```diff
@@ -33,6 +37,13 @@ export async function format(
     fix = false;
   }
 
+  // If the project has a .clang-format – use it. Else use the default as a
```
jinwoo: Nit. The last `a` -> `an`.

jinwoo:
Some tests are failing though.

ofrobots:
Tests fixed.

Issue #144, opened 2018-04-03
Generated package.json has inconsistent main


ofrobots:
See https://github.com/google/ts-style/pull/143/files#r178920063.

* [ ] We should default to `build/src/index.js`.
* [ ] We should include a `types` property in the generated `package.json`.

PR #179, opened 2018-06-22
fix(init): init package with better main, types

ofrobots:
Fixes: https://github.com/google/ts-style/issues/144

### DIFF
--- a/src/init.ts
+++ b/src/init.ts
@@ -229,12 +229,8 @@ export async function init(options: Options): Promise<boolean> {
       return false;
     }
 
-    try {
-      packageJson = DEFUALT_PACKAGE_JSON;
-      generatedPackageJson = true;
-    } catch (err2) {
-      throw err2;
-    }
+    packageJson = DEFUALT_PACKAGE_JSON;
+    generatedPackageJson = true;
   }
 
   const addedDeps = await addDependencies(packageJson, options);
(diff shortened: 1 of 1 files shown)


src/init.ts
```diff
@@ -210,17 +230,16 @@ export async function init(options: Options): Promise<boolean> {
     }
 
     try {
-      // TODO(ofrobots): add proper error handling.
-      cp.spawnSync('npm', ['init', '-y']);
-      packageJson = await readJson('./package.json');
+      packageJson = DEFUALT_PACKAGE_JSON;
```
kjin: Do we still need the try-catch here?

src/init.ts
```diff
@@ -21,7 +21,19 @@ import * as path from 'path';
 import {Options} from './cli';
 import {readFilep as read, readJsonp as readJson, writeFileAtomicp as write} from './util';
 
-const pkg = require('../../package.json') as PackageJson;
+const pkg = require('../../package.json');
```
kjin: Why did the type annotation here disappear?

ofrobots: That wasn't an annotation but rather a cast. The cast was not necessary.

ofrobots: Also – we don't use this yet – but going forward we would stronger types for package.json available from new resolveJsonModule in TS 2.9.

src/init.ts
ofrobots: Indeed not. Removed.

PR #182, opened 2018-06-27
Add more tests

ofrobots:
In an effort to improve coverage, I started working some tests. It turns out that our current coverage numbers are a lie as `init.ts` wasn't accounted for. It only gets tested via `test-kitchen.ts` which is hard to report for coverage purposes (it uses an tarball-installed version of the package rather than the existing files).

Here some more tests. Still more to come, but no reason to sit on these. These can land.

### DIFF
--- a/src/init.ts
+++ b/src/init.ts
@@ -45,13 +45,13 @@ export interface PackageJson {
   version?: string;
   devDependencies?: Bag<string>;
   scripts?: Bag<string>;
-  name: string;
-  description: string;
-  main: string;
-  types: string;
-  files: string[];
-  license: string;
-  keywords: string[];
+  name?: string;
+  description?: string;
+  main?: string;
+  types?: string;
+  files?: string[];
+  license?: string;
+  keywords?: string[];
 }
 
 async function query(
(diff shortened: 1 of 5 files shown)


src/init.ts
```diff
@@ -45,13 +45,13 @@ export interface PackageJson {
...
   devDependencies?: Bag<string>;
   scripts?: Bag<string>;
-  name: string;
-  description: string;
-  main: string;
-  types: string;
-  files: string[];
-  license: string;
-  keywords: string[];
+  name?: string;
```
JustinBeckwith: May be that I'm missing context - why are these all optional now?

test/test-format.ts
```diff
@@ -37,7 +37,7 @@ const OPTIONS: Options = {
   logger: {log: console.log, error: console.error, dir: nop}
 };
 
-test.serial('format should return false for well-formatted files', t => {
+test.serial('format should return true for well-formatted files', t => {
```
JustinBeckwith: Well that's confusing :) Was the title of the test just wrong before?

ofrobots: true

src/init.ts
ofrobots: They should always have been optional. The schema for `package.json` isn't very strong and there is no requirement that all of these keys must be present.

PR #183, opened 2018-07-03
fix: disable esModuleInterop

ofrobots:
BREAKING CHANGE: esModuleInterop causes synthetic imports to leak from
libraries to their users. Users must compile with allowSyntheticImports
as well if they use a library that uses this flag.

This has been causing a lot of churn in libraries that use gts.

### DIFF
--- a/tsconfig-google.json
+++ b/tsconfig-google.json
@@ -4,7 +4,6 @@
     "allowUnreachableCode": false,
     "allowUnusedLabels": false,
     "declaration": true,
-    "esModuleInterop": true,
     "forceConsistentCasingInFileNames": true,
     "lib": ["es2016"],
     "module": "commonjs",
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -15,8 +15,8 @@
  * limitations under the License.
  */
 import * as path from 'path';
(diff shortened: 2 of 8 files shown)


tsconfig-google.json
```diff
@@ -4,7 +4,6 @@
     "allowUnreachableCode": false,
     "allowUnusedLabels": false,
     "declaration": true,
-    "esModuleInterop": true,
```
JustinBeckwith: Do we need to disable synthetic default imports as well?  Or am I getting my compiler flag wires crossed?

ofrobots: No.. esModuleInterop implicitly enables synthetic imports.  See here: https://www.typescriptlang.org/docs/handbook/compiler-options.html

JustinBeckwith: We explicitly turn it on here tho: https://github.com/google/ts-style/pull/183/files#diff-19e170b73889957f2d2c69de5a7a5d7aL3

ofrobots: D'oh.

ofrobots: #184

Issue #64, opened 2017-10-19
`gts clean` doesn't follow tsconfig.json dependency chain


kjin:
If my `tsconfig.json` extends `tsconfig.other.json` with no additional parameters, `gts clean` will fail because it can't read `build` in `tsconfig.json`.

It should try to detect the extension chain and find the `build` field in `tsconfig.other.json`. Any other command that relies on a field in `tsconfig.json` might have the same issue.

ofrobots:
This wasn't an intended use-case (I was imagining a single hop). I am concerned that we may have to do these dependency walks in quite a few cases. This will need an abstraction to ensure that this gets handled transparently everywhere. If a PR is available, I think we should consider this feature request.

PR #185, opened 2018-07-09
fix: allow 'gts clean' to follow tsconfig.json dependency chain

jsamar17:
I wrote and edited the tests to show the new functionality and to catch circular references. This still should be abstracted to effect all commands that look at the contents of tsconfig.json. Also there is no support for compileOnSave yet. fix: #64

### DIFF
--- a/src/clean.ts
+++ b/src/clean.ts
@@ -30,7 +30,6 @@ export async function clean(options: Options): Promise<boolean> {
   const tsconfig = (await getTSConfig(options.targetRootDir)) as TSConfig;
   if (tsconfig.compilerOptions && tsconfig.compilerOptions.outDir) {
     const outDir = tsconfig.compilerOptions.outDir;
-
     if (outDir === '.') {
       options.logger.error(
           `${chalk.red('ERROR:')} ${chalk.gray('compilerOptions.outDir')} ` +
--- a/src/clean.ts
+++ b/src/clean.ts
@@ -30,6 +30,7 @@ export async function clean(options: Options): Promise<boolean> {
   const tsconfig = (await getTSConfig(options.targetRootDir)) as TSConfig;
   if (tsconfig.compilerOptions && tsconfig.compilerOptions.outDir) {
     const outDir = tsconfig.compilerOptions.outDir;
+
     if (outDir === '.') {
       options.logger.error(
           `${chalk.red('ERROR:')} ${chalk.gray('compilerOptions.outDir')} ` +
--- a/src/util.ts
+++ b/src/util.ts
@@ -13,37 +13,104 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-
+ 
 import * as fs from 'fs';
 import * as path from 'path';
 import pify from 'pify';
 import rimraf from 'rimraf';
-
+ 
 export const readFilep = pify(fs.readFile);
 export const rimrafp = pify(rimraf);
 export const writeFileAtomicp = pify(require('write-file-atomic'));
-
+ 
 export async function readJsonp(jsonPath: string) {
   return JSON.parse(await readFilep(jsonPath));
 }
-
+ 
 export interface ReadFileP {
-  (path: string, encoding: string): Promise<string>;
+  (path: string, encoding: string): Promise < string > ;
 }
-
+ 
 export function nop() {
   /* empty */
 }
-
+ 
 /**
  * Find the tsconfig.json, read it, and return parsed contents.
  * @param rootDir Directory where the tsconfig.json should be found.
  */
+
+ // Old getTSConfig function
+// export async function getTSConfig(
+//   rootDir: string, customReadFilep ? : ReadFileP): Promise < {} > {
+//   const tsconfigPath = path.join(rootDir, 'tsconfig.json');
+//   //console.log(tsconfigPath);
+//   customReadFilep = customReadFilep || readFilep;
+//   const json = await customReadFilep(tsconfigPath, 'utf8');
+//   console.log(json);
+//   const contents = JSON.parse(json);
+//   return contents;
+// }
+ 
 export async function getTSConfig(
-    rootDir: string, customReadFilep?: ReadFileP): Promise<{}> {
+  rootDir: string, customReadFilep ? : ReadFileP): Promise < {} > {
+ 
   const tsconfigPath = path.join(rootDir, 'tsconfig.json');
+
+  //console.log(tsconfigPath);
   customReadFilep = customReadFilep || readFilep;
   const json = await customReadFilep(tsconfigPath, 'utf8');
-  const contents = JSON.parse(json);
+  //console.log(json);
+  let contents = JSON.parse(json);
+
+  if (contents["extends"]) {
+    const thing = await getExtension(contents["extends"]);
+    contents =combineTSConfig(thing, contents);
+ 
+  }
   return contents;
 }
+ 
+
+async function getExtension(
+  filePath: string, customReadFilep ? : ReadFileP): Promise < {} > {
+  customReadFilep = customReadFilep || readFilep;
+  const json = await customReadFilep(filePath, 'utf8');
+  let contents = JSON.parse(json);
+ 
+  if (contents["extends"]) {
+    const nextFile = await getExtension(contents["extends"]);
+    contents =combineTSConfig(nextFile, contents);
+  }
+  //console.log(contents);
+  return contents;
+ 
+}
+ 
+function combineTSConfig(
+ ts1: any, ts2: any): {}{
+// const TSProperties = ["compilerOptions", "files", "include", "exclude"];
+let result = {"compilerOptions" : {},
+  "files": [],
+  "include":[],
+  "exclude": []
+};
+
+
+ts1.compilerOptions = ts1.compilerOptions || {};
+ts1.files = ts1.files || [];
+ts1.include = ts1.include || [];
+ts1.exclude = ts1.exclude || [];
+ts2.compilerOptions = ts2.compilerOptions || {};
+ts2.files = ts2.files || [];
+ts2.include = ts2.include || [];
+ts2.exclude = ts2.exclude || [];
+
+
+Object.assign(result.compilerOptions, ts1.compilerOptions, ts2.compilerOptions);
+result.files = result.files.concat(ts1.files, ts2.files);
+result.include = result.include.concat(ts1.include, ts2.include);
+result.exclude = result.exclude.concat(ts1.exclude, ts2.exclude);
+
+return result;
+ }
--- a/src/util.ts
+++ b/src/util.ts
@@ -13,34 +13,35 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
- 
+
 import * as fs from 'fs';
+import {access, read} from 'fs/promises';
 import * as path from 'path';
 import pify from 'pify';
 import rimraf from 'rimraf';
- 
+
 export const readFilep = pify(fs.readFile);
 export const rimrafp = pify(rimraf);
 export const writeFileAtomicp = pify(require('write-file-atomic'));
- 
+
 export async function readJsonp(jsonPath: string) {
   return JSON.parse(await readFilep(jsonPath));
 }
- 
+
 export interface ReadFileP {
-  (path: string, encoding: string): Promise < string > ;
+  (path: string, encoding: string): Promise<string>;
 }
- 
+
 export function nop() {
   /* empty */
 }
- 
+
 /**
  * Find the tsconfig.json, read it, and return parsed contents.
  * @param rootDir Directory where the tsconfig.json should be found.
  */
 
- // Old getTSConfig function
+// Old getTSConfig function
 // export async function getTSConfig(
 //   rootDir: string, customReadFilep ? : ReadFileP): Promise < {} > {
 //   const tsconfigPath = path.join(rootDir, 'tsconfig.json');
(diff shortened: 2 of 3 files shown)


src/util.ts
```diff
@@ -38,12 +39,77 @@ export function nop() {
...
+    contents = combineTSConfig(extension, contents);
+  }
+  return Promise.resolve(contents);
 }
+
+async function getExtension(
+    filePath: string, customReadFilep: ReadFileP,
+    readFiles: string[]): Promise<ConfigFile> {
+  customReadFilep = customReadFilep || readFilep;
+  if (!filePath) {
```
kjin: This check isn't necessary.

kjin: Discussed offline as to why. In general, typechecking allows us to forgo the check as long as the function's not user-facing. (User in this case is anyone who imports this file.)

src/clean.ts
```diff
@@ -30,6 +30,7 @@ export async function clean(options: Options): Promise<boolean> {
   const tsconfig = (await getTSConfig(options.targetRootDir)) as TSConfig;
   if (tsconfig.compilerOptions && tsconfig.compilerOptions.outDir) {
     const outDir = tsconfig.compilerOptions.outDir;
+
```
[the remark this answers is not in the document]
jsamar17: 👍

src/util.ts
```diff
@@ -15,6 +15,7 @@
  */
 
 import * as fs from 'fs';
+import {access, read} from 'fs/promises';
```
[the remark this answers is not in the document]
jsamar17: 👍

src/util.ts
```diff
@@ -38,12 +39,77 @@ export function nop() {
...
+    readFiles: string[]): Promise<ConfigFile> {
+  customReadFilep = customReadFilep || readFilep;
+  if (!filePath) {
+    throw new Error('Undefined passed in');
+  }
+  if (readFiles.indexOf(filePath) !== -1) {
+    // did I throw the error correctly?
+    throw new Error('Circular Reference Detected');
+  }
+  readFiles.push(filePath);
```
kjin: `filePath` may be relative to the location of `tsconfig.json`. This might result in issues if say, `tsconfig.json` depends on `foo/tsconfig.bar.json` which depends on `tsconfig.json`: does that last file refer to `./tsconfig.json` or `./foo/tsconfig.json`?

src/util.ts
```diff
@@ -38,12 +39,77 @@ export function nop() {
...
+  const readArr = ['tsconfig.json'];
+  if (contents['extends']) {
+    const extension =
+        await getExtension(contents['extends'], customReadFilep, readArr);
+    contents = combineTSConfig(extension, contents);
+  }
+  return Promise.resolve(contents);
 }
+
+async function getExtension(
```
kjin: I would consider calling this `getBase` instead, because the the config file at `filePath` overrides the config that it extends, not the other way around.

src/util.ts
```diff
@@ -38,12 +39,77 @@ export function nop() {
...
+  }
+  // console.log(contents);
+  return Promise.resolve(contents);
+}
+
+// the inherited config file overwrites the base config file's "files",
+// "include", and "exclude" fields and combines compiler options
+function combineTSConfig(base: ConfigFile, inherited: ConfigFile): ConfigFile {
+  // const TSProperties = ["compilerOptions", "files", "include", "exclude"];
+  const result = {
```
kjin: I think we can just initialize to `{ compilerOptions: {} }`. If neither `base` nor `inherited` has any of the other fields defined, then an empty `files` (for example) might be semantically different than an undefined `files`.

src/util.ts
```diff
@@ -38,12 +39,77 @@ export function nop() {
...
+    contents = combineTSConfig(nextFile, contents);
+  }
+  // console.log(contents);
+  return Promise.resolve(contents);
+}
+
+// the inherited config file overwrites the base config file's "files",
+// "include", and "exclude" fields and combines compiler options
+function combineTSConfig(base: ConfigFile, inherited: ConfigFile): ConfigFile {
+  // const TSProperties = ["compilerOptions", "files", "include", "exclude"];
```
kjin: Could you remove this line?

src/util.ts
```diff
@@ -38,12 +39,77 @@ export function nop() {
...
+  if (contents['extends']) {
+    const nextFile =
+        await getExtension(contents['extends'], customReadFilep, readFiles);
+    contents = combineTSConfig(nextFile, contents);
+  }
+  // console.log(contents);
+  return Promise.resolve(contents);
+}
+
+// the inherited config file overwrites the base config file's "files",
```
kjin: Can you move this comment to the lines for which it's relevant? Which, as of this version is around 110.

test/test-util.ts
```diff
@@ -31,4 +31,74 @@ test('get should parse the correct tsconfig file', async t => {
   t.deepEqual(contents, FAKE_CONFIG);
 });
 
+
+
+// TODO: test for error due to circular reference
```
kjin: This TODO can be removed now, right?

test/test-util.ts
```diff
@@ -31,4 +31,74 @@ test('get should parse the correct tsconfig file', async t => {
...
+test('should throw an error if it finds a circular reference', async t => {
+  const FAKE_DIRECTORY = '/some/fake/directory';
+  const FAKE_CONFIG = {files: ['b'], extends: 'CONFIG1'};
+  const CONFIG1 = {extends: 'CONFIG2'};
```
kjin: I would suggest calling these FAKE_CONFIG2, etc.

test/test-util.ts
```diff
@@ -31,4 +31,74 @@ test('get should parse the correct tsconfig file', async t => {
...
+
+
+test('should follow dependency chain caused by extends files', async t => {
+  const FAKE_DIRECTORY = '/some/fake/directory';
+  const FAKE_CONFIG = {
+    compilerOptions: {compileExtraFastPleas: 'SuperSpeed'},
+    files: ['b'],
+    extends: 'CONFIG1'
+  };
+  const CONFIG1 = {include: ['/stuff/*'], extends: 'CONFIG2'};
```
kjin: Can you add `exclude` and `compilerOptions` to demonstrate that (1) overwriting happens correctly?

test/test-util.ts
```diff
@@ -31,4 +31,74 @@ test('get should parse the correct tsconfig file', async t => {
...
+        return Promise.reject('File Not Found');
+    }
+  }
+  const contents = await getTSConfig(FAKE_DIRECTORY, fakeReadFilep);
+  t.deepEqual(contents, combinedConfig);
+});
+
+
+
+// TODO: test for circular reference error with symlinks
```
kjin: I think we don't necessarily need to test this if other comments are addressed. This is because even with symlinks, we are keeping a cache of "seen" config files, and having symlinks doesn't change the fact that a file will eventually be "seen" twice if it's part of a cycle.

src/util.ts
```diff
@@ -38,12 +39,77 @@ export function nop() {
...
   const tsconfigPath = path.join(rootDir, 'tsconfig.json');
+
   customReadFilep = customReadFilep || readFilep;
   const json = await customReadFilep(tsconfigPath, 'utf8');
-  const contents = JSON.parse(json);
-  return contents;
+  let contents = JSON.parse(json);
+
+  const readArr = ['tsconfig.json'];
+  if (contents['extends']) {
```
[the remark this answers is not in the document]
jsamar17: 👍

src/util.ts
```diff
@@ -38,12 +39,77 @@ export function nop() {
 /**
  * Find the tsconfig.json, read it, and return parsed contents.
  * @param rootDir Directory where the tsconfig.json should be found.
+ * If the tsconfig.json file has an "extends" field hop down the dependency tree until it ends
```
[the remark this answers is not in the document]
jsamar17: 👍

src/util.ts
```diff
@@ -38,12 +39,77 @@ export function nop() {
...
-  return contents;
+  let contents = JSON.parse(json);
+
+  const readArr = ['tsconfig.json'];
+  if (contents['extends']) {
+    const extension =
+        await getExtension(contents['extends'], customReadFilep, readArr);
+    contents = combineTSConfig(extension, contents);
+  }
+  return Promise.resolve(contents);
```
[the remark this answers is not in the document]
jsamar17: 👍

src/util.ts
jsamar17: 👍

src/util.ts
jsamar17: 👍

src/util.ts
jsamar17: 👍

test/test-util.ts
jsamar17: 👍

src/util.ts
```diff
@@ -38,12 +39,77 @@ export function nop() {
 /**
  * Find the tsconfig.json, read it, and return parsed contents.
  * @param rootDir Directory where the tsconfig.json should be found.
+ * If the tsconfig.json file has an "extends" field hop down the dependency tree until it ends
  */
+
+
+export interface ConfigFile {
```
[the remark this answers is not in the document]
jsamar17: 👍

src/util.ts
```diff
@@ -38,12 +39,77 @@ export function nop() {
...
+
+async function getExtension(
+    filePath: string, customReadFilep: ReadFileP,
+    readFiles: string[]): Promise<ConfigFile> {
+  customReadFilep = customReadFilep || readFilep;
+  if (!filePath) {
+    throw new Error('Undefined passed in');
+  }
+  if (readFiles.indexOf(filePath) !== -1) {
+    // did I throw the error correctly?
```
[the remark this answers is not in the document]
jsamar17: 👍

src/util.ts
jsamar17: 👍

src/util.ts
jsamar17: 👍

test/test-util.ts
jsamar17: 👍

src/util.ts
```diff
@@ -15,10 +15,10 @@
  */
 
 import * as fs from 'fs';
-import {access, read} from 'fs/promises';
 import * as path from 'path';
 import pify from 'pify';
 import rimraf from 'rimraf';
+import {file} from 'tmp';
```
kjin: Because `file` is so broad in its meaning, I would suggest `import * as tmp` so as to be more specific. `tmp.file(...)` is more meaningful than `file(...)`

src/util.ts
```diff
@@ -39,77 +39,73 @@ export function nop() {
...
-    readFiles: string[]): Promise<ConfigFile> {
-  customReadFilep = customReadFilep || readFilep;
-  if (!filePath) {
-    throw new Error('Undefined passed in');
-  }
   if (readFiles.indexOf(filePath) !== -1) {
-    // did I throw the error correctly?
-    throw new Error('Circular Reference Detected');
+    throw new Error(
+        'Circular reference in ' + readFiles[readFiles.indexOf(filePath)]);
```
kjin: I suggest using the following string interpolation syntax:
```
const x = 'hello'
console.log(`${x}, world`) // prints "hello, world"
```

src/util.ts
```diff
@@ -39,77 +39,73 @@ export function nop() {
...
       result.compilerOptions, base.compilerOptions, inherited.compilerOptions);
   delete result.extends;
   return result;
-}
\ No newline at end of file
+}
+/**
+ * An interface containing the high level data fields present in Config Files
+ */
+
```
kjin: Nit: Remove empty line between docs for X and X itself.

src/util.ts
```diff
@@ -39,77 +39,73 @@ export function nop() {
...
+  const result = {'compilerOptions': {}, 'extends': {}};
 
   Object.assign(result, base, inherited);
   Object.assign(
       result.compilerOptions, base.compilerOptions, inherited.compilerOptions);
   delete result.extends;
   return result;
-}
\ No newline at end of file
+}
```
kjin: Nit: Add a line between a class/interface definition and the docs for the next definition. (I think this is also true of `getBase`)

src/util.ts
```diff
@@ -39,77 +39,73 @@ export function nop() {
...
-  // console.log(contents);
-  return Promise.resolve(contents);
+
+  return contents;
 }
 
-// the inherited config file overwrites the base config file's "files",
-// "include", and "exclude" fields and combines compiler options
+/**
+ * takes in 2 config files
```
kjin: Can you write full sentences in JSDocs? Here as well as in `@param` parameter descriptions.

src/util.ts
```diff
@@ -39,77 +39,73 @@ export function nop() {
...
 function combineTSConfig(base: ConfigFile, inherited: ConfigFile): ConfigFile {
-  // const TSProperties = ["compilerOptions", "files", "include", "exclude"];
-  const result = {
-    'compilerOptions': {},
-    'files': [],
-    'include': [],
-    'exclude': [],
-    'extends': {}
-  };
+  const result = {'compilerOptions': {}, 'extends': {}};
```
kjin: Why is `extends` one of the fields here? It seems like it's guaranteed to be deleted later.

kjin: If it's there for type definition purposes, I would annotate result to be of type `ConfigFile`.

src/clean.ts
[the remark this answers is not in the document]
kjin: Unless I'm mistaken, it seems like this file wasn't reverted to the original.

src/util.ts
```diff
@@ -39,77 +39,73 @@ export function nop() {
...
 export async function getTSConfig(
     rootDir: string, customReadFilep?: ReadFileP): Promise<ConfigFile> {
+  customReadFilep = customReadFilep || readFilep;
   // array to hold accessed files, used to check for circular references
-
-
-
-  const tsconfigPath = path.join(rootDir, 'tsconfig.json');
-
+  const readArr = [''];
```
kjin: Consider making this a `Set`.

test/test-util.ts
```diff
@@ -97,8 +97,86 @@ test('should follow dependency chain caused by extends files', async t => {
 });
 
 
+test('should throw an error if it finds a circular reference', async t => {
```
kjin: Is this the same test as above?

test/test-util.ts
```diff
@@ -97,8 +97,86 @@ test('should follow dependency chain caused by extends files', async t => {
...
+          default:
+            return Promise.reject('File Not Found');
+        }
+      }
+      const contents = await getTSConfig(FAKE_DIRECTORY, fakeReadFilep);
+      t.deepEqual(contents, combinedConfig);
+    });
+
+test(
+    'When a file, all filepaths should be relative to the config file currently being read',
```
kjin: The first part of this sentence seems to be missing a predicate?

test/test-util.ts
```diff
@@ -97,8 +97,86 @@ test('should follow dependency chain caused by extends files', async t => {
...
+        return Promise.reject('File Not Found');
+    }
+  }
+  await t.throws(
+      getTSConfig(FAKE_DIRECTORY, fakeReadFilep), Error,
+      'Circular Reference Detected');
+});
+
+test(
+    'when a file contains an extends field, the base file is loaded first then overridden by the inherited files',
```
kjin: This seems like it is already covered in a test above. (`should follow dependency chain caused by extends files`)

src/util.ts
jsamar17: This module isn't used in this file so I removed it

src/clean.ts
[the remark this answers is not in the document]
eyqs: test

test/test-util.ts
kjin: I see, makes sense.

src/util.ts
jsamar17: If combineTSConfig is called that means that a file was contains an extends field. If the extends field is not removed the program will create an infinite loop. I updated it so that it is now part of the ConfigFile definition.

kjin:
cc/ @jplaisted

test/test-util.ts
jsamar17: yes, I removed the redundant test

src/util.ts
```diff
@@ -38,12 +38,72 @@ export function nop() {
...
+ */
+async function getBase(
+    filePath: string, customReadFilep: ReadFileP, readFiles: Set<string>,
+    currentDir: string): Promise<ConfigFile> {
+  customReadFilep = customReadFilep || readFilep;
+
+  filePath = path.resolve(currentDir, filePath);
+
+  if (readFiles.has(filePath)) {
+    throw new Error('Circular reference in ${filePath}');
```
kjin: For posterity, can you include a comment on why we throw here (rather than give up and return immediately)?

src/util.ts
```diff
@@ -38,12 +38,72 @@ export function nop() {
...
   return contents;
 }
+
+/**
+ * Takes in 2 config files
+ * @param base is loaded first
+ * @param inherited is then loaded and overwrites base
+ */
+function combineTSConfig(base: ConfigFile, inherited: ConfigFile): ConfigFile {
+  const result: ConfigFile = {'compilerOptions': {}};
```
kjin: Nit: Remove `''` around `compilerOptions`.

src/util.ts
```diff
@@ -38,12 +38,72 @@ export function nop() {
...
+
+  Object.assign(result, base, inherited);
+  Object.assign(
+      result.compilerOptions, base.compilerOptions, inherited.compilerOptions);
+  delete result.extends;
+  return result;
+}
+
+/**
+ * An interface containing the high level data fields present in Config Files
```
kjin: Nit: `top level` is more characteristic here.

test/test-util.ts
```diff
@@ -20,15 +20,132 @@ import {getTSConfig} from '../src/util';
...
+    include: ['/stuff/*'],
+    exclude: ['doesnt/look/like/anything/to/me']
+  };
+
+
+  function fakeReadFilep(
+      configPath: string, encoding: string): Promise<string> {
+    switch (configPath) {
+      case '/some/fake/directory/FAKE_CONFIG1':
+
```
kjin: I think I missed this before. What is this case for? Is this falling through to the next case?

(Here and also in the next test)

test/test-util.ts
```diff
@@ -20,15 +20,132 @@ import {getTSConfig} from '../src/util';
...
+          default:
+            return Promise.reject('File Not Found');
+        }
+      }
+      const contents = await getTSConfig(FAKE_DIRECTORY, fakeReadFilep);
+      t.deepEqual(contents, combinedConfig);
+    });
+
+test(
+    'When reading a file, all filepaths should be relative to the config file currently being read',
```
kjin: Nit: Make this lowercase.

test/test-util.ts
jsamar17: Similar to a symlink the file FAKE_CONFIG1 has 2 addresses /some/fake/directory/tsconfig.json and /some/fake/directory/FAKE_CONFIG1. I could edit this out and make the circular references point back to /some/fake/directory/tsconfig.json I kept this in all the tests for consistancy but I can remove it

test/test-util.ts
```diff
@@ -16,19 +16,121 @@
 import test from 'ava';
 import * as path from 'path';
 
-import {getTSConfig} from '../src/util';
+import {ConfigFile, getTSConfig} from '../src/util';
+
+
```
kjin: Can you make a one line break here instead of three.

test/test-util.ts
```diff
@@ -16,19 +16,121 @@
...
+
+/**
+ * Creates a fake promisified readFile function from a map
+ * @param myMap contains a filepath as the key and a ConfigFile object as the
+ * value. The fake promisified readFile function has the same interface of
+ * fs.readFile and takes in a
+ * @param configPath and returns a corresponding ConfigFile object or throws an
```
kjin: This `@param` annotation is a little misleading, since it's a parameter of the returned function, not this one. I think it's acceptable to just say that the return value has the same interface as a promisifed fs.readFile function.

PR #186, opened 2018-07-10
feat: print clang-format output 

### LINKED ISSUE #181: Print `clang-format` output
In #60:

> A fix for this should be landed shortly (#62); all it does is add an error message if linting succeeds and formatting fails. I'm planning to add parsing `clang-format` XML output as a follow-up (PRs from your end would be more than welcome too). That being said, it's best to run `gts fix` to fix format issues anyway; I believe `clang-format` output is not necessarily meant to be human-interpretable.

When `clang-format` fails, all we do is print that there were formatting issues. `gts fix` fixes those issues, but it would still be nice to show the lines of code that need to be addressed.

lee-tammy:
Still working on the feature so that it can support non-utf8 characters.

Fixes #181

### DIFF
--- a/package.json
+++ b/package.json
@@ -42,13 +42,16 @@
   "dependencies": {
     "chalk": "^2.4.1",
     "clang-format": "1.2.3",
+    "entities": "^1.1.1",
     "inquirer": "^6.0.0",
     "meow": "^5.0.0",
     "pify": "^3.0.0",
     "rimraf": "^2.6.2",
     "tslint": "^5.9.1",
     "update-notifier": "^2.5.0",
-    "write-file-atomic": "^2.3.0"
+    "utfstring": "^2.0.0",
+    "write-file-atomic": "^2.3.0",
+    "xml2js": "^0.4.19"
   },
   "devDependencies": {
     "@types/glob": "^5.0.35",
--- a/package.json
+++ b/package.json
@@ -52,6 +52,8 @@
     "write-file-atomic": "^2.3.0"
   },
   "devDependencies": {
+    "@types/diff": "^3.5.1",
+    "@types/entities": "^1.1.0",
     "@types/glob": "^5.0.35",
     "@types/inquirer": "^0.0.42",
     "@types/make-dir": "^1.0.3",
--- a/package.json
+++ b/package.json
@@ -49,9 +49,7 @@
     "rimraf": "^2.6.2",
     "tslint": "^5.9.1",
     "update-notifier": "^2.5.0",
-    "utfstring": "^2.0.0",
-    "write-file-atomic": "^2.3.0",
-    "xml2js": "^0.4.19"
+    "write-file-atomic": "^2.3.0"
   },
   "devDependencies": {
     "@types/glob": "^5.0.35",
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "gts",
-  "version": "0.7.1",
+  "version": "0.8.0",
   "description": "Google TypeScript Style",
   "repository": "google/ts-style",
   "main": "build/src/index.js",
--- a/package.json
+++ b/package.json
@@ -42,6 +42,7 @@
   "dependencies": {
     "chalk": "^2.4.1",
     "clang-format": "1.2.3",
+    "diff": "^3.5.0",
     "entities": "^1.1.1",
     "inquirer": "^6.0.0",
     "meow": "^5.0.0",
--- a/src/format.ts
+++ b/src/format.ts
@@ -18,7 +18,8 @@ import * as path from 'path';
 import {Options} from './cli';
 import {createProgram} from './lint';
 
-const clangFormat = require('clang-format');
+// Exported for testing purposes.
+export const clangFormat = require('clang-format');
 
 const BASE_ARGS_FILE = ['-style=file'];
 const BASE_ARGS_INLINE =
@@ -56,7 +57,7 @@ export async function format(
       program.getRootFileNames().filter(f => !f.endsWith('.d.ts'));
 
   if (fix) {
-    return await fixFormat(srcFiles, baseClangFormatArgs);
+    return fixFormat(srcFiles, baseClangFormatArgs);
   } else {
     const result = await checkFormat(srcFiles, baseClangFormatArgs);
     if (!result) {
--- a/src/format.ts
+++ b/src/format.ts
@@ -15,7 +15,6 @@
  */
 import chalk from 'chalk';
 import * as jsdiff from 'diff';
-import {resolve} from 'dns';
 import * as entities from 'entities';
 import * as fs from 'fs';
 import * as path from 'path';
@@ -181,7 +180,7 @@ async function getDiffObj(
 }
 
 /**
- * Performs formatting fixes to the original string 🌷
+ * Performs formatting fixes to the original string
  *
  * @param data original string
  * @param errOffset start index of the formatting issue
--- a/src/format.ts
+++ b/src/format.ts
@@ -15,9 +15,10 @@
  */
 import * as fs from 'fs';
 import * as path from 'path';
+import {promisify} from 'util';
+
 import {Options} from './cli';
 import {createProgram} from './lint';
-import { promisify } from 'util';
 
 // Exported for testing purposes.
 export const clangFormat = require('clang-format');
@@ -39,7 +40,6 @@ const BASE_ARGS_INLINE =
  */
 export async function format(
     options: Options, files: string[] = [], fix = false): Promise<boolean> {
-    
   if (options.dryRun && fix) {
     options.logger.log('format: skipping auto fix since --dry-run was passed');
     fix = false;
@@ -99,7 +99,8 @@ function fixFormat(srcFiles: string[], baseArgs: string[]): Promise<boolean> {
  *
  * @param srcFiles list of source files
  */
-function checkFormat(srcFiles: string[], baseArgs: string[], options: Options): Promise<boolean> {
+function checkFormat(srcFiles: string[], baseArgs: string[], options: Options):
+    Promise<boolean> {
   return new Promise<boolean>((resolve, reject) => {
     let output = '';
     const arrOffset: number[] = [];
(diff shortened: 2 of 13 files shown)


lee-tammy:
Should work on files with non-utf8 characters now

src/format.ts
```diff
@@ -110,8 +120,117 @@ function checkFormat(srcFiles: string[], baseArgs: string[]): Promise<boolean> {
...
+      return;
+    }
+
+    for (let j = 0; j < formatErr.fix.length; j++) {
+      formatErr.fix[j] = entities.decodeXML(formatErr.fix[j]);
+    }
+
+    const read = promisify(fs.readFile);
+    const argNum = 3;
+    const file = process.argv[argNum + i - 1];
```
kjin: Is it guaranteed that the files will be in `process.argv`? I believe the list of arguments passed to `clang-format` come from a different source?

kjin: Also, could you add an inline comment explaining why it's safe to use `i` to index into a different array than `files`?

src/format.ts
```diff
@@ -110,8 +120,117 @@ function checkFormat(srcFiles: string[], baseArgs: string[]): Promise<boolean> {
...
+    const text = await read(file, 'utf8');
+
+    const fixed =
+        performFixes(text, formatErr.offset, formatErr.length, formatErr.fix);
+    const diff = jsdiff.structuredPatch(
+        'oldFile', 'newFile', text, fixed, 'old', 'new', {context: 3});
+    jsdiff.applyPatch('diff', diff);
+    printDiffs(file, diff.hunks, options);
```
kjin: Suggestion: The name of this function `findFormatErrorLines` doesn't suggest that lines will get printed. Rather than printing here, consider returning the necessary information for printing, and call `printDiffs` externally from `findFormatErrorLines`.

src/format.ts
```diff
@@ -110,8 +120,117 @@ function checkFormat(srcFiles: string[], baseArgs: string[]): Promise<boolean> {
...
+
+    const read = promisify(fs.readFile);
+    const argNum = 3;
+    const file = process.argv[argNum + i - 1];
+    const text = await read(file, 'utf8');
+
+    const fixed =
+        performFixes(text, formatErr.offset, formatErr.length, formatErr.fix);
+    const diff = jsdiff.structuredPatch(
+        'oldFile', 'newFile', text, fixed, 'old', 'new', {context: 3});
```
kjin: Do any of these fields indicate the name of the file?

src/format.ts
```diff
@@ -110,8 +120,117 @@ function checkFormat(srcFiles: string[], baseArgs: string[]): Promise<boolean> {
...
+    jsdiff.applyPatch('diff', diff);
+    printDiffs(file, diff.hunks, options);
+  }
+}
+
+/**
+ * Performs formatting fixes to the original string 🌷
+ *
+ * @param data original string
+ * @param errOffset
```
kjin: Could you fill in descriptions for the missing `@param` fields?

src/format.ts
```diff
@@ -110,8 +120,117 @@ function checkFormat(srcFiles: string[], baseArgs: string[]): Promise<boolean> {
...
+ * @param errLength
+ * @param replacements
+ */
+function performFixes(
+    data: string, errOffset: string[], errLength: string[],
+    replacements: string[]) {
+  const replaced: string[] = [];
+  replaced.push(substring(data, 0, +errOffset[0]));
```
kjin: Instead of using the `+` operator here to cast offsets and lengths from strings to numbers, could you perform the cast when you first assign the field?

kjin: Also, there are multiple ways to cast from strings to numbers... does `+` fit our use case here?

src/format.ts
```diff
@@ -110,8 +120,117 @@ function checkFormat(srcFiles: string[], baseArgs: string[]): Promise<boolean> {
...
+  for (let i = 0; i < errOffset.length - 1; i++) {
+    replaced.push(replacements[i]);
+    replaced.push(
+        substring(data, +errOffset[i] + +errLength[i], +errOffset[i + 1]));
+  }
+  replaced.push(replacements[errOffset.length - 1]);
+  replaced.push(substring(
+      data, +errOffset[errOffset.length - 1] + +errLength[errOffset.length - 1],
+      Buffer.byteLength(data, 'utf8')));
```
kjin: 👍

src/format.ts
```diff
@@ -110,8 +120,117 @@ function checkFormat(srcFiles: string[], baseArgs: string[]): Promise<boolean> {
...
+}
+
+/**
+ * Prints the lines with formatting issues
+ *
+ * @param file
+ * @param diffs contains all information about the formatting changes
+ * @param options
+ */
+function printDiffs(file: string, diffs: any, options: Options) {
```
kjin: I believe that if you `import` from `diff`, then you can get the interface that matches the `any` here.

src/format.ts
```diff
@@ -15,11 +15,18 @@
...
 import * as path from 'path';
+import {promisify} from 'util';
+
 import {Options} from './cli';
 import {createProgram} from './lint';
 
 // Exported for testing purposes.
 export const clangFormat = require('clang-format');
+const xml2js = require('xml2js');
+const chalk = require('chalk');
```
kjin: Can you use `import * as x from 'x'` instead of `const x = require('x')`? TS will prompt you for type definitions -- you can get them with

`npm install --save-dev @types/x`

The `@types/` packages contain type definitions for each package. Excluding xml and utf (because you're not using them), I believe the rest of them all have this associated package.

src/format.ts
```diff
@@ -110,8 +120,117 @@ function checkFormat(srcFiles: string[], baseArgs: string[]): Promise<boolean> {
...
+      }
     });
+    options.logger.log('\n');
   });
 }
+
+/**
+ * Substring by bytes
+ * 
+ * @param str
```
kjin: Can you fill in the parameter decsriptions here?

src/format.ts
```diff
@@ -110,8 +120,117 @@ function checkFormat(srcFiles: string[], baseArgs: string[]): Promise<boolean> {
...
+ * @param indexStart 
+ * @param indexEnd 
+ * @param encoding 
+ */
+function substring(
+    str: string, indexStart: number, indexEnd: number, encoding = 'utf8') {
+  return Buffer.from(str, encoding)
+      .slice(indexStart, indexEnd)
+      .toString(encoding);
+}
```
kjin: Can you add a new line to the end of this file? (It's our general practice to do so)

package.json
```diff
@@ -42,13 +42,16 @@
...
     "clang-format": "1.2.3",
+    "entities": "^1.1.1",
     "inquirer": "^6.0.0",
     "meow": "^5.0.0",
     "pify": "^3.0.0",
     "rimraf": "^2.6.2",
     "tslint": "^5.9.1",
     "update-notifier": "^2.5.0",
-    "write-file-atomic": "^2.3.0"
+    "utfstring": "^2.0.0",
```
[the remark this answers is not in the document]
lee-tammy: done

src/format.ts
```diff
@@ -38,7 +45,7 @@ export async function format(
     fix = false;
   }
 
-  // If the project has a .clang-format – use it. Else use the default as an
+  // If the project has a .clang-format i– use it. Else use the default as an
```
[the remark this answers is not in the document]
lee-tammy: done

src/format.ts
```diff
@@ -15,11 +15,18 @@
  */
 import * as fs from 'fs';
 import * as path from 'path';
+import {promisify} from 'util';
```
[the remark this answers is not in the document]
lee-tammy: done

src/format.ts
```diff
@@ -110,8 +120,117 @@ function checkFormat(srcFiles: string[], baseArgs: string[]): Promise<boolean> {
...
+    if (formatErr.length === null || formatErr.offset === null ||
+        formatErr.fix === null) {
+      return;
+    }
+
+    for (let j = 0; j < formatErr.fix.length; j++) {
+      formatErr.fix[j] = entities.decodeXML(formatErr.fix[j]);
+    }
+
+    const read = promisify(fs.readFile);
```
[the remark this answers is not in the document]
lee-tammy: done

src/format.ts
```diff
@@ -110,8 +120,117 @@ function checkFormat(srcFiles: string[], baseArgs: string[]): Promise<boolean> {
...
+  for (let i = 1; i < files.length; i++) {
+    const formatErr = {
+      offset: output.match(/(?<=offset=\')(\d+)(?=\')/g),
+      length: output.match(/(?<=length=\')(\d+)(?=\')/g),
+      fix: output.match(/(?<=length=\'\d+\'>)(.*)(?=<\/replacement>)/g)
+    };
+
+    if (formatErr.length === null || formatErr.offset === null ||
+        formatErr.fix === null) {
+      return;
```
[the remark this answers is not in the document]
lee-tammy: So, instead of checking if the variables are null, should I check if they are not null and using continue ...and have an else block with the print statement that tells the user that there are no formatting issues? The only case that I can think of where formatErr.length, offset, and fix would be null, is if there are no formatting issues. 

```
if (formatErr.length !== null && formatErr.offset !== null &&
    formatErr.fix !== null) {
  continue;
}else{
  options.logger.log('clang-format reported no errors')
}
```

src/format.ts
lee-tammy: done

src/format.ts
lee-tammy: I just realized that too... I used process.argv because it provides the file name, but my algorithm makes the assumption that the user passes the absolute path of a file. The xml string does not provide the file names, so how can I get them?

src/format.ts
lee-tammy: done

src/format.ts
lee-tammy: done

src/format.ts
lee-tammy: done. Now it is casting when properties of formatting error object are first assigned.

src/format.ts
lee-tammy: done

src/format.ts
lee-tammy: done

src/format.ts
lee-tammy: fixed!

src/format.ts
lee-tammy: done

src/format.ts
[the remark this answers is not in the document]
lee-tammy: Take a look at the changes I made so far. I'm returning an empty string[] instead of undefined. I don't think it's necessary to print to stderr because it is not an error when formatErr.length/offset/fix is null. it means that the format of the file is fine and does not need to be changed.

package.json
```diff
@@ -42,6 +42,7 @@
   "dependencies": {
     "chalk": "^2.4.1",
     "clang-format": "1.2.3",
+    "entities": "^1.1.1",
```
ofrobots: Did you miss `diff` here?

src/format.ts
```diff
@@ -110,8 +122,127 @@ function checkFormat(srcFiles: string[], baseArgs: string[]): Promise<boolean> {
...
+async function getDiffObj(
+    file: string, replacements: replacement[]): Promise<jsdiff.IUniDiff> {
+  const text = await readFilep(file, 'utf8');
+
+  const fixed = performFixes(text, replacements);
+  const diff = jsdiff.structuredPatch(
+      file, 'no new file', text, fixed, 'no old header', 'no new header',
+      {context: 3});
+  jsdiff.applyPatch('diff', diff);
+  return Promise.resolve(diff);
```
ofrobots: Change this to simply `return diff`. Async functions automatically wrap return values in a promise.

src/format.ts
```diff
@@ -110,8 +122,127 @@ function checkFormat(srcFiles: string[], baseArgs: string[]): Promise<boolean> {
...
+ * Substring using bytes
+ *
+ * @param str original string
+ * @param indexStart position where to start extraction
+ * @param indexEnd position (up to, but not including) where to end extraction
+ * @param encoding
+ */
+function substring(
+    str: string, indexStart: number, indexEnd: number, encoding = 'utf8') {
+  return Buffer.from(str, encoding)
```
ofrobots: This looks odd. Since we already had a string – we already had decoded it using whatever encoding that was necessary. Why are we converting to a buffer and then converting back to a string? This does extra work in encoding/decoding, and will be buggy in cases where a different encoding ought to have been used (as rare as it might be).

ofrobots:
I just tried this out. This looks really cool!

image

src/format.ts
lee-tammy: done

package.json
lee-tammy: yes, added it back

src/format.ts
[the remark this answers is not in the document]
kjin: > The only case that I can think of where formatErr.length, offset, and fix would be null, is if there are no formatting issues.

Yep, I forgot that it's possible for there to not be replacements at all, indicating no formatting issues. For handling this, you can consider `continue`-ing earlier, even before the regexes are being executed, by checking whether the `replacement` tag occurs at all. That way, when you get to the point where `formatErr.length === null || formatErr.offset === null || formatErr.fix === null` is being evaluated, you know that if that statement is true, it indicates that `clang-format` didn't emit expected output. Even though I wouldn't expect `clang-format` to change its output xml format in the future, this would help us disambiguate whether we are continuing because there are no formatting issues, or because an unexpected change occurred in `clang-format`'s output.

src/format.ts
```diff
@@ -13,20 +13,22 @@
...
 import * as fs from 'fs';
 import * as path from 'path';
+
 import {Options} from './cli';
 import {createProgram} from './lint';
-import { promisify } from 'util';
-import { start } from 'repl';
+import {readFilep} from './util';
+
+interface replacement {
```
kjin: Interfaces names should be capitalized. Also, could you add a JSDoc describing the role of this object? (By JSDoc I mean `/** */` style rather than `//` style)

src/format.ts
```diff
@@ -100,11 +101,11 @@ function fixFormat(srcFiles: string[], baseArgs: string[]): Promise<boolean> {
  *
  * @param srcFiles list of source files
  */
-function checkFormat(srcFiles: string[], baseArgs: string[], options: Options): Promise<boolean> {
+async function checkFormat(
```
kjin: I believe this function does not need to be `async`, since the use of `await` is in a nested function.

src/format.ts
```diff
@@ -115,101 +116,132 @@ function checkFormat(srcFiles: string[], baseArgs: string[], options: Options):
...
-    
-    out.on('end', () => {
-      findFormatErrorLines(output, options)
-        .then(() => {
-          resolve(output.indexOf('<replacement ') === -1 ? true : false);
-        });
+
+    out.on('end', async () => {
+      const files = output.split('<?xml version=\'1.0\'?>\n');
+      for (let i = 1; i < files.length; i++) {
```
kjin: I would recommend using [`slice`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/slice) so that we can just start this loop at `i = 0`.

src/format.ts
```diff
@@ -115,101 +116,132 @@ function checkFormat(srcFiles: string[], baseArgs: string[], options: Options):
...
+        const replacements = getReplacements(files[i]);
+        if (replacements.length > 0) {
+          const diff = await getDiffObj(srcFiles[i - 1], replacements)
```
kjin: This line seems to be missing a `;`? Maybe you need to run `npm run fix`

src/format.ts
```diff
@@ -115,101 +116,132 @@ function checkFormat(srcFiles: string[], baseArgs: string[], options: Options):
...
+ *
+ * @param output xml string from clangFormat
  */
-async function findFormatErrorLines(output: string, options: Options){
-  const files = output.split('<?xml version=\'1.0\'?>\n');
-
-  for(let i = 1; i < files.length; i++){
+function getReplacements(output: string): replacement[] {
+  const replacements: replacement[] = [];
+  const offsets: string[]|null = output.match(/(?<=offset=\')(\d+)(?=\')/g);
```
kjin: I think the approach in this function would benefit from an inline comment.

src/format.ts
kjin: You can probably just assign empty strings, if you aren't reading them elsewhere.

src/format.ts
kjin: There isn't an available `bytesSubstring` function out there. I was thinking that we should publish this as separate module (it does in fact have subtle differences from `substring`).

Given that this function is only used for showing output, and that JS files are _generally_ written with UTF-8 encoding, I feel like it's acceptable to say (for now) that if your file is encoded differently, you're on your own. That being said, any static tool that prints code snippets is going to have to do the same operation somewhere, so maybe we should look to `tslint` or others for inspiration. I wouldn't be too surprised if `tslint` makes the UTF-8 assumption as well.

kjin: Discovered this: https://github.com/palantir/tslint/issues/2810

src/format.ts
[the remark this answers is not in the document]
lee-tammy: done

src/format.ts
lee-tammy: done

src/format.ts
lee-tammy: oops, forgot to remove

src/format.ts
lee-tammy: done

src/format.ts
lee-tammy: added

src/format.ts
lee-tammy: done. Check to see if it's ok

kjin:
cc/ @jplaisted

src/format.ts
```diff
@@ -92,7 +107,8 @@ function fixFormat(srcFiles: string[], baseArgs: string[]): Promise<boolean> {
  *
  * @param srcFiles list of source files
  */
-function checkFormat(srcFiles: string[], baseArgs: string[]): Promise<boolean> {
+function checkFormat(srcFiles: string[], baseArgs: string[], options: Options):
```
kjin: Nit: Can we emulate `format` and put the `options` argument first?

src/format.ts
```diff
@@ -110,8 +126,137 @@ function checkFormat(srcFiles: string[], baseArgs: string[]): Promise<boolean> {
...
+ * returns all of the necessary replacements for the file
+ *
+ * @param output xml string from clangFormat
+ */
+function getReplacements(fileXML: string): Replacement[] {
+  const replacements: Replacement[] = [];
+
+  let xmlLines = fileXML.split('\n');
+  // The first and last two elements in xmlLines are not needed
+  xmlLines = xmlLines.slice(1, xmlLines.length - 2);
```
kjin: Suggestion: It wasn't immediately obvious to me that `xmlLines.length - 2` is correct. If you define `xmlLines` as `fileXML.trim().split('\n')` then it will guarantee one element per non-empty line, so you could probably use `xmlLines.length - 1`.

src/format.ts
```diff
@@ -110,8 +126,137 @@ function checkFormat(srcFiles: string[], baseArgs: string[]): Promise<boolean> {
...
+
+  for (let i = 0; i < xmlLines.length; i++) {
+    // Uses regex to capture the xml attribute and element
+    // XML format: <replacement offset='OFFSET' length='LENGTH'>FIX</replacement
```
kjin: Nit: Missing ending `>`

src/format.ts
```diff
@@ -110,8 +126,137 @@ function checkFormat(srcFiles: string[], baseArgs: string[]): Promise<boolean> {
...
+  return replaced.join('');
+}
+
+/**
+ * Prints the lines with formatting issues
+ *
+ * @param diffs contains all information about the formatting changes
+ * @param options
+ */
+function printDiffs(diffs: jsdiff.IUniDiff, options: Options) {
```
kjin: Suggestion: For string constructions in this file, try string interpolation.

```
const x = 2;
console.log(`${x + x} = 4`) // prints "4 = 4"

File: test/test-format.ts [hunk 28]
```diff
@@ -18,15 +18,18 @@ import test from 'ava';
...
+import {Logger, Options} from '../src/cli';
 import * as format from '../src/format';
 import {nop} from '../src/util';
 
 import {withFixtures} from './fixtures';
 
 // clang-format won't pass this code because of trailing spaces.
-const BAD_CODE = 'export const foo = [ 2 ];';
-const GOOD_CODE = 'export const foo = [2];';
+const BAD_CODE = '//🦄\nexport const foo = [ \"2\" ];';
```
kjin: As we discussed in person, let's have the unicorn be in its own test.

src/format.ts
lee-tammy: done

src/format.ts
lee-tammy: done

src/format.ts
lee-tammy: added

test/test-format.ts
lee-tammy: done

src/format.ts
lee-tammy: I'm using string interpolation now. Check if it's okay

src/format.ts
```diff
@@ -154,13 +154,14 @@ function checkFormat(srcFiles: string[], baseArgs: string[], options: Options):
...
-  // The first and last two elements in xmlLines are not needed
-  xmlLines = xmlLines.slice(1, xmlLines.length - 2);
+  let xmlLines = fileXML.trim().split('\n');
+  // first and last elements are outer 'replacements' tags
+  xmlLines = xmlLines.slice(1, xmlLines.length - 1);
 
   for (let i = 0; i < xmlLines.length; i++) {
     // Uses regex to capture the xml attribute and element
-    // XML format: <replacement offset='OFFSET' length='LENGTH'>FIX</replacement
+    // XML format: <replacement offset='OFFSET'
```
kjin: This looks like it was "fixed" with npm run fix. Can you put the `<...>` on its own line so it doesn't get broken up?

Issue #81, opened 2017-11-12
Discussion: `gts watch`


JustinBeckwith:
As we convert libraries from gulp to gts, the only thing I'm really missing is `npm run watch`.  I'd like to run that command, have it automatically watch for changes in `*.ts*` files or `package.json`, and react by running the 'compile` and `test` commands.  I could see a syntax like this:

`$ gts watch`

By default, it would monitor the sources directory for changes, and run `npm run compile` on change.  We wouldn't let users change it.  After compilation, it would run `npm test`.  On test failure, we would have to be careful to not let the process stop.  

If we want to get fancy - we could support a syntax that runs the start process instead of tests on file change.  This would be useful for web applications where you're more concerned with live development:

`$ gts watch --start`

I'm a little conflicted, because this goes beyond the scope of what we're building - but it is a useful thing I just want to have out of the box with most TypeScript projects.  

Thoughts?

ofrobots:
`tsc --watch` already exists. It doesn't integrate with tests, but perhaps a good starting point. WDYT?

JustinBeckwith:
There are a few things I don't like about `tsc --watch`:
- On file change, I want `npm run compile` to be run, so I don't have to worry about the tsc command being the same in two places. 
- If I'm writing a library, I almost always want the tests afterward
- If I'm working on an app, I almost always want to restart my process, and maybe even reload the web page I'm working on

`tsc --watch` solves a problem in a vacuum, but rarely is actually enough in real life (IMO).

carnesen:
I feel this. I've been finding myself on these `@google-cloud` projects running `tsc --watch` in one shell and `mocha build/test --watch` in another, and I've given up on maintaining lint/format correctness in "real time" as I develop. (Instead I fix once at the end before submitting the PR). I assume here you're asking for a tool that helps to orchestrate those processes. As a linting/formatting tool though, `compile` and `test` are probably outside `gts`s scope.

ofrobots:
Thanks! I think I understand the use-case. PTAL at the WIP PR #207, and let me know if this would address the use-case.

JustinBeckwith:
As time has gone on, I now think this is a bad idea :)

PR #207, opened 2018-09-21
feat(watch): add a watch verb

ofrobots:
Fixes: https://github.com/google/ts-style/issues/81

WIP but opening now to get early feedback.  

Still to do:
- [ ] implement tests
- [ ] child process error handling
- [ ] avoid running a then/else command when a previous one has not completed.


src/watch.ts
```diff
@@ -0,0 +1,71 @@
...
+    }
+    return ret;
+  };
+
+  ts.createWatchProgram(host);
+  return true;
+}
+
+function reportDiagnostic(diagnostic: ts.Diagnostic) {
+  console.error(
```
kjin: Use `options.logger` here?

carnesen:
I see now that this project is more of a multi-purpose ts dev cli, not limited to just style and formatting 👍 . From the description of what the `watch` command does, it would be more consistent with other CLI tools to put that functionality under `gts compile --watch`... I would expect `gts watch` just to watch. So if I wanted to compile and test on change I'd do `gts watch --then 'npm run compile && npm run test`. Or maybe `gts watch --on-change 'npm run compile && npm run test`.

ofrobots:
> I would expect gts watch just to watch

I am not sure if I agree with this. AFAIU `tsc -w` doesn't just watch, it runs forever and triggers incremental recompilation. Ideally `tsc` should have some hooks to run commands after each {un}successful incremental compilation, but it doesn't. `gts` is stepping in to provide this missing functionality.

> So if I wanted to compile and test on change I'd do gts watch --then 'npm run compile && npm run test.

This would be suboptimal. We would want an *incremental* compilation when using `watch`. `npm run compile` will force a full compilation. Right now the only way to cause incremental compilation (AFAICT) is programmatically. We need a `gts` verb to do incremental compilation, and IMO `watch` matches my intuition the most here.

carnesen:
@ofrobots I don't want to say your intuition is wrong, but I'd argue it's inconsistent with the other CLI APIs in this space (ava, tsc, mocha, tslint, eslint, jest), for which the user provides a `--watch` flag to enable an incremental/watcher mode. If the proposed "watch" command's primary job is to do incremental compilation, `gts compile --watch` would be more intuitive to more users. As you said, that's `tsc`'s API! 

If it's just `gts watch`, it's unclear to the user whether that's enabling a watch mode for `fix`, `check`, `compile` or something else. As I mentioned previously, I'd actually guess that a `watch` command would actually do just and only that (watch) similar to [chokidar-cli](https://github.com/kimmobrunfeldt/chokidar-cli).

ofrobots:
I'm going to abandon this until I can get my head around a clear value proposition.

PR #259, opened 2019-02-05
feat!: switch to prettier for formatting

ofrobots:
NOTE: this targets the next branch.

BREAKING CHANGE: Switch to prettier as the code formatter.

### DIFF
--- a/prettier.config.js
+++ b/prettier.config.js
@@ -1,9 +1,4 @@
 module.exports = {
-  bracketSpacing: false,
-  printWidth: 80,
-  semi: true,
   singleQuote: true,
-  tabWidth: 2,
   trailingComma: 'es5',
-  useTabs: false,
 };
--- a/prettier.config.js
+++ b/prettier.config.js
@@ -1,0 +1,9 @@
+module.exports = {
+  bracketSpacing: false,
+  printWidth: 80,
+  semi: true,
+  singleQuote: true,
+  tabWidth: 2,
+  trailingComma: 'es5',
+  useTabs: false,
+};
--- a/src/format.ts
+++ b/src/format.ts
@@ -82,9 +82,16 @@ async function checkFormat(srcFiles: string[], options: Options, fix: boolean) {
   const checks = configs.map(({ file, config }: FileConfig) => {
     config = config || PRETTIER_OPTIONS;
     const contents = fs.readFileSync(file, 'utf8');
-    const formatted = prettier.format(contents, config!);
-    const wellFormatted = contents === formatted;
-    if (wellFormatted) {
+
+    let formatted: string;
+    try {
+      formatted = prettier.format(contents, config!);
+    } catch (e) {
+      options.logger.log(`${file}: ${e}`);
+      return false;
+    }
+
+    if (contents === formatted) {
       return true;
     }
     if (fix) {
--- a/src/format.ts
+++ b/src/format.ts
@@ -45,18 +45,27 @@ export async function format(
       ? files
       : program.getRootFileNames().filter(f => !f.endsWith('.d.ts'));
 
-  if (fix) {
-    fixFormat(srcFiles);
-    return true;
-  } else {
-    const result = await checkFormat(srcFiles, options);
-    if (!result) {
-      options.logger.log(
-        'prettier reported errors... run `gts fix` to address.'
-      );
-    }
-    return result;
+  const result = await checkFormat(srcFiles, options, fix);
+  if (result === false) {
+    options.logger.log(
+      'prettier reported errors... run `gts fix` to address.'
+    );
   }
+  return result;
+}
+
+interface FileConfig {
+  file: string;
+  config: prettier.Options|null;
+}
+
+async function mapFilesToFileConfigs(srcFiles: string[]): Promise<FileConfig[]> {
+  const configs = await Promise.all(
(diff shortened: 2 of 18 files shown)


prettier.config.js
```diff
@@ -0,0 +1,9 @@
+module.exports = {
+  bracketSpacing: false,
```
JustinBeckwith: I know the point here isn't to paint the bikeshed, but I personally like bracket spacing *ducks*

prettier.config.js
```diff
@@ -0,0 +1,9 @@
...
+  printWidth: 80,
```
JustinBeckwith: printWidth, semi, tabWidth, and useTabs are all set to their defaults here:
https://prettier.io/docs/en/options.html

Should we just omit those?

src/format.ts
```diff
@@ -75,217 +42,57 @@ export async function format(
...
-    const length: string[] | null = /length='(\d+)'/g.exec(xmlLines[i]);
-    const fix: string[] | null = /length='\d+'>(.*)<\/replacement>/g.exec(
-      xmlLines[i]
-    );
-
-    if (length === null || offset === null || fix === null) {
-      throw new Error('Unable to get replacement');
+async function checkFormat(srcFiles: string[], options: Options) {
+  const configs = await Promise.all(srcFiles.map(file =>
+      prettier.resolveConfig(file)));
```
JustinBeckwith: One concern I have here is the # of different formats prettier supports for config files:
https://prettier.io/docs/en/configuration.html

Does using this method mean that we'll only drop a `prettier.config.js` if none of the 4 options are present?  Or is there potential for us to clobber on init?

prettier.config.js
ofrobots: If you look at the delta, this is ported from the `.prettierrc` that you added.

Is there rationalization for the value you used?

prettier.config.js
ofrobots: There are here because you added them to `.prettierrc`. I actually prefer sticking to the defaults  as much as possible.

src/format.ts
ofrobots: This should look for any of the variants of the config files that `prettier` supports. If none is found, it will use the default from `gts`.

prettier.config.js
JustinBeckwith: I had just copied over what's used in all of the other libraries 🤷‍♂️   I doubt those choices were really scrutinized.  I don't have a strong opinion here, just a preference for defaults when nobody otherwise cares. What do you think?

prettier.config.js
JustinBeckwith: +1 to sticking to the defaults.  Single quotes are likely the only one that would be hella disruptive.

JustinBeckwith:
Not sure if on purpose or a typo - why the `!` after `feat`?

ofrobots:
https://github.com/conventional-commits/conventionalcommits.org/issues/43

JustinBeckwith:
TIL!  That's an awesome idea folks.  I have no forking clue which of my commits today are breaking changes.

src/format.ts
ofrobots: Added a test for this so that you don't have to take my word for it.

ofrobots:
@google/google-node-team any opinions on the default prettier config?

prettier.config.js
ofrobots: I've updated the config to a minimal set (IMO). PTAL.

src/format.ts
```diff
@@ -45,55 +45,56 @@ export async function format(
...
-  } else {
-    const result = await checkFormat(srcFiles, options);
-    if (!result) {
-      options.logger.log(
-        'prettier reported errors... run `gts fix` to address.'
-      );
-    }
-    return result;
+  const result = await checkFormat(srcFiles, options, fix);
+  if (result === false) {
```
kjin: Why not just `!result`?

src/format.ts
```diff
@@ -45,55 +45,56 @@ export async function format(
...
+  const configs = await mapFilesToFileConfigs(srcFiles);
 
-  const checks = srcFiles.map((file, index) => {
-    const config = configs[index] || PRETTIER_OPTIONS;
-    config.filepath = config.filepath || file;
+  const checks = configs.map(({file, config}: FileConfig) => {
+    config = config || PRETTIER_OPTIONS;
     const contents = fs.readFileSync(file, 'utf8');
-    const formatted = prettier.format(contents, config);
+    const formatted = prettier.format(contents, config!);
```
kjin: I think the `!` isn't necessary here?

ofrobots: It is. The compiler is not smart enough to figure this out.

ofrobots:
PTAL.

src/format.ts
kjin: Interesting. Curiously, the following compiles https://gist.github.com/kjin/db2b50cc2ee1fbdd34404a2c092b917a

test/test-format.ts
```diff
@@ -290,3 +290,30 @@ test.serial(
     );
   }
 );
+
+// Files that cannot be formatted should be left untouched.
+test.serial('format should leave the kitty unharmed', t => {
+  const KITTY = `
+   /\\**/\\
+  ( o_o  )_)
```
kjin: 🐱

PR #271, opened 2019-02-17
fix: specify prettier parser

### LINKED ISSUE #270: Prettier doesn't seem to be parsing TypeScript
I'm trying to get https://github.com/googleapis/nodejs-spanner converted over the master branch of `gts` both to a.) address #267 and b.) take the `1.x` pre release out for a spin :) After getting through a few other issues that I filed in previous bugs, I'm getting errors like this:

```
No parser and no filepath given, using 'babel' the parser now but this will throw an error in the future. Please specify a parser or a filepath so one can be inferred.
/Users/beckwith/Code/nodejs-spanner/test/transaction-runner.ts: SyntaxError: Unexpected token (89:19)
  87 | 
  88 |       ExtendedRunner = class ExtendedRunner extends Runner {
> 89 |         protected async _run(transaction): Promise<void> {
     |                   ^
  90 |           return runFn(transaction);
  91 |         }
  92 |       };
No parser and no filepath given, using 'babel' the parser now but this will throw an error in the future. Please specify a parser or a filepath so one can be inferred.
No parser and no filepath given, using 'babel' the parser now but this will throw an error in the future. Please specify a parser or a filepath so one can be inferred.
/Users/beckwith/Code/nodejs-spanner/system-test/spanner.ts: SyntaxError: Unexpected token (19:15)
  17 | import * as assert from 'assert';
  18 | // tslint:disable-next-line
> 19 | import pLimit = require('p-limit');
     |               ^
  20 | import concat = require('concat-stream');
  21 | import * as crypto from 'crypto';
  22 | import * as extend from 'extend';
```

It kinda looks like `prettier` doesn't know that it should be parsing TypeScript, and that it's getting confused by TypeScript syntax.

JustinBeckwith:
fixes #270

### DIFF
--- a/src/format.ts
+++ b/src/format.ts
@@ -80,7 +80,10 @@ async function checkFormat(srcFiles: string[], options: Options, fix: boolean) {
   const configs = await mapFilesToFileConfigs(srcFiles);
 
   const checks = configs.map(({ file, config }: FileConfig) => {
-    config = config || PRETTIER_OPTIONS;
+    config = Object.assign(
+      { parser: 'typescript' },
+      config || PRETTIER_OPTIONS
+    );
     const contents = fs.readFileSync(file, 'utf8');
 
     let formatted: string;


ofrobots:
I would prefer #273.

JustinBeckwith:
Better fix :)

PR #283, opened 2019-02-26
refactor(test): convert tests to mocha

### LINKED ISSUE #260: Tracking issue: Convert ava to mocha in tests
- [ ] test-clean.ts
- [ ] test-format.ts
- [ ] test-init.ts
- [ ] test-kitchen.ts
- [ ] test-lint.ts
- [ ] test-util.ts

JustinBeckwith:
Fixes #260.  @ofrobots I did most of the grunt work here, and now I'm just fighting a few failing tests.  I pushed this to upstream so we could maybe collaborate on this one.   

Side note: really suggest enabling 'hide whitespace changes' for this one

### DIFF
--- a/test/test-format.ts
+++ b/test/test-format.ts
@@ -44,7 +44,7 @@ describe('format', () => {
     logger: { log: console.log, error: console.error, dir: nop },
   };
 
-  it('format should return true for well-formatted files', async () => {
+  it('format should return true for well-formatted files', () => {
     return withFixtures(
       {
         'tsconfig.json': JSON.stringify({ files: ['a.ts'] }),
@@ -52,7 +52,7 @@ describe('format', () => {
       },
       async () => {
         const result = await format.format(OPTIONS, [], false);
-        assert.ok(result);
+        assert.strictEqual(result, true);
       }
     );
   });
@@ -65,7 +65,7 @@ describe('format', () => {
       },
       async () => {
         const result = await format.format(OPTIONS, [], false);
-        assert.ok(!result);
+        assert.strictEqual(result, false);
       }
     );
   });
@@ -79,7 +79,7 @@ describe('format', () => {
       },
       async () => {
         const result = await format.format(OPTIONS, [], false);
-        assert.ok(result);
+        assert.strictEqual(result, true);
       }
     );
   });
@@ -92,7 +92,7 @@ describe('format', () => {
       },
       async fixturesDir => {
         const result = await format.format(OPTIONS, [], true);
-        assert.ok(result);
+        assert.strictEqual(result, true);
         const contents = fs.readFileSync(
           path.join(fixturesDir, 'a.ts'),
           'utf8'
@@ -111,7 +111,7 @@ describe('format', () => {
       },
       async () => {
         const okay = await format.format(OPTIONS);
-        assert.ok(okay);
+        assert.strictEqual(okay, true);
       }
     );
   });
@@ -125,7 +125,7 @@ describe('format', () => {
       },
       async () => {
         const okay = await format.format(OPTIONS);
-        assert.ok(!okay);
+        assert.strictEqual(okay, false);
       }
     );
   });
@@ -139,7 +139,7 @@ describe('format', () => {
       },
       async fixturesDir => {
         const okay = await format.format(OPTIONS, [], true);
-        assert.ok(okay);
+        assert.strictEqual(okay, true);
         const contents = fs.readFileSync(
           path.join(fixturesDir, 'a.ts'),
           'utf8'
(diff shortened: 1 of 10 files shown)


ofrobots:
Awesome! I will take a look when I get a chance.

vitkarpov:
Out of curiosity, why Mocha and not Jest? 😄

JustinBeckwith:
@vitkarpov honestly, just consistency.  We use `mocha` for everything else, so this just makes it easier to not think and get stuff done.

vitkarpov:
Gotcha, it's totally fine, I'm just asking 😉

vitkarpov:
@JustinBeckwith which codemode are you using to apply this transition?

JustinBeckwith:
There were some really nasty merge conflicts on this one, so please bear with me.  🐻

JustinBeckwith:
@google/google-node-team this is ready for a look!

test/test-format.ts
```diff
@@ -132,116 +139,119 @@ test.serial(
...
+  it('format globs listed in include', () => {
+    return withFixtures(
+      {
+        'tsconfig.json': JSON.stringify({ include: ['dirb/*'] }),
+        dira: { 'a.ts': GOOD_CODE },
+        dirb: { 'b.ts': BAD_CODE },
+      },
+      async () => {
+        const okay = await format.format(OPTIONS);
+        assert.strictEqual(okay, false);
```
kjin: Nit -- I see both `assert.strictEqual(okay, false)` and `assert.ok(!okay)`, can we just use one of those?

test/test-init.ts
```diff
@@ -58,124 +56,126 @@ function hasExpectedDependencies(packageJson: PackageJson): boolean {
...
-    },
-    async () => {
-      const result = await init.init(OPTIONS_YARN);
-      t.truthy(result);
-
-      const contents = await readJson('./package.json');
-      t.truthy(contents.scripts.prepare === 'yarn run compile');
-    }
-  );
+  it('init should support yarn', async () => {
```
kjin: Nit: this method has `async () => { return Promise }`, I've also seen `async () => { await Promise }` and `() => { return Promise }`, they're all valid but should we stick with just one pattern?

test/test-init.ts
```diff
@@ -58,124 +56,126 @@ function hasExpectedDependencies(packageJson: PackageJson): boolean {
...
+      {
+        'package.json': JSON.stringify({ name: 'test' }),
+        'yarn.lock': '',
+      },
+      async () => {
+        const result = await init.init(OPTIONS_YARN);
+        assert.ok(result);
+
+        const contents = await readJson('./package.json');
+        assert.ok(contents.scripts.prepare === 'yarn run compile');
```
kjin: Can we use `strictEqual` here?

test/test-kitchen.ts
```diff
@@ -66,140 +62,138 @@ const execOpts = {
...
- * CLEAN UP - remove the staging directory when done.
- */
-test.after.always('cleanup staging', async () => {
-  if (!keep) {
-    stagingDir.removeCallback();
-  }
+  /**
+   * CLEAN UP - remove the staging directory when done.
+   */
+  after('cleanup staging', () => {
```
kjin: Nit: Should `after` be right after `before`? I think that's our "standard" in other tests. Could go either way though

test/test-lint.ts
```diff
@@ -25,103 +25,108 @@ import { nop } from '../src/util';
...
-});
+  it('lint should return true on good code', () => {
+    return withFixtures(
+      {
+        'tsconfig.json': JSON.stringify({ files: ['a.ts'] }),
+        'a.ts': GOOD_CODE,
+      },
+      async () => {
+        const okay = lint.lint(OPTIONS);
+        assert.strictEqual(okay, true);
```
kjin: Here and elsewhere, use `assert.ok`?

test/test-format.ts
JustinBeckwith: yeah that's gross, good feedback

test/test-lint.ts
JustinBeckwith: I went the other way.  I prefer to use `assert.strictEqual` to verify the object returned is actually a `bool`, and not a random object or something.  I made this consistent throughout.

JustinBeckwith:
@kjin this is ready for another look!

test/test-lint.ts
kjin: We already get guarantees on return types from TypeScript, IMO we should not duplicate that check in our tests. But this is fine, the outcome is the same.

PR #285, opened 2019-02-28
Support Yarn

### LINKED ISSUE #281: What's about yarn?
Hey there!

Thanks for this project. I'm wondering would it be useful to allow `gts` using yarn instead of npm while init?

Why did I pay attention to it? I use yarn across my projects and for a new project I created package.json first (using `yarn init`), added `gts` as a dev dependency and run `npx gts init`. What I got was package-lock.json near yarn.lock 🤔  

They have pretty much the same API, so I just changed [npm here](https://github.com/google/gts/blob/master/src/init.ts#L295) to yarn and that's it. I'd suggest to automatically look at "is there `yarn.lock` or not" and switch between them (not only for init but everywhere, probably an adaptor needs to be introduced so it's not a trivia)

What do you, fellows, think about it? Probably, there's a reason to use `npm` only?

vitkarpov:
Fixes #281

### DIFF
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -62,7 +62,7 @@ const cli = meow({
     --help        Prints this help message.
     -y, --yes     Assume a yes answer for every prompt.
     -n, --no      Assume a no answer for every prompt.
-    --dry-run     Don't make any acutal changes.
+    --dry-run     Don't make any actual changes.
     --yarn        Use yarn instead of npm.
 
 	Examples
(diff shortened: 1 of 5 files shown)


vitkarpov:
I signed it!

ofrobots:
This looks great, thanks for the contribution :).

vitkarpov:
Thank for having me 😉 It's my pleasure 🎉

PR #294, opened 2019-03-09
fix: Remove manual package.json interface

grant:
- Removes manual `package.json` types.
  - Uses `@types/package-json` instead.

Note: Import order sorted automatically via VSCode.


JustinBeckwith:
Thanks for the fix!

PR #295, opened 2019-03-11
chore(deps): update make-dir to 2.1.0

vitkarpov:
`make-dir@2.1.0` has its own types inside the package, so `@types/make-dir` is redundant now.
There's a problem with typings in 2.1.0 though, you can't use `import * as makeDir from 'make-dir'` anymore since typings say it doesn't export a function.
Let's use default import instead.

Details: https://github.com/sindresorhus/make-dir/issues/12


vitkarpov:
No problem 👍

PR #296, opened 2019-03-11
fix!: lint enforce trailing commas

ofrobots:
While not part of the internal style guide (yet), enforcing trailing
commas make sense as they make diffs simpler. They would would give us
better congruence with the JS style guide.

Fixes: https://github.com/google/gts/issues/292


ofrobots:
> do we have this as the default for the prettier config?

Here: https://github.com/google/gts/blob/master/prettier.config.js#L3

PR #305, opened 2019-04-01
append .cmd to npm/yarn command when in Windows

### LINKED ISSUE #278: `npx gts init` command does not install gts
I followed the instruction on the top page (run `npx gts init`) but kept getting error when I open my .ts file on Atom. I finally resolved the problem by executing `npm install gts` in my project directory.

A probable cause may be that my project directory only has a "src" sub-directory and nothing else (no "node_modules" sub-directory or "package.json" file. Please investigate if this is an issue and fix the code.

dan1wang:
fix #278 

Got a bunch of error when running mocha test on Windows...

### DIFF
--- a/src/init.ts
+++ b/src/init.ts
@@ -281,8 +281,10 @@ export async function init(options: Options): Promise<boolean> {
   if (!options.dryRun) {
     // --ignore-scripts so that compilation doesn't happen because there's no
     // source files yet.
+    let command = getPkgManagerName(options.yarn);
+    if (process.platform === 'win32') command += '.cmd';
     cp.spawnSync(
-      getPkgManagerName(options.yarn),
+      command,
       ['install', '--ignore-scripts'],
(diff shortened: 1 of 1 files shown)


src/init.ts
```diff
@@ -281,8 +281,10 @@ export async function init(options: Options): Promise<boolean> {
   if (!options.dryRun) {
     // --ignore-scripts so that compilation doesn't happen because there's no
     // source files yet.
+    let command = getPkgManagerName(options.yarn);
+    if (process.platform === 'win32') command += '.cmd';
```
vitkarpov: I'd suggest to untangle it a little bit, and make it as API for `getPkgManagerName`, like:

```
getPkgManagerName(options.yarn, process.platform)
```

This logic (with `cmd` or without one) would be encapsulated and also easily covered with unit tests. What do you think?

PR #306, opened 2019-04-02
Fix npx command on Windows and make sure build/test work on Windows too

### LINKED ISSUE #278: `npx gts init` command does not install gts

dan1wang:
Fix #278 (`npx gts init` doesn't work on Windows)
Fix test files so all build/test will pass on Windows
- renamep(tarball, `${stagingPath}/gts.tgz`) fails on Winodws. Rename first and then move (use fs-extra module)
- on Windows, when you pass a directory to `getTSConfig()`, the directory gets resolved with "D:\" prefix (or something similar). Resolve all directories first in test files so assertion won't fail.
- Windows converts `"\n"` to `"\r\n"`,
- Misc. changes for Windows and to make Prettier happy

CI result on Azure DevOps https://dev.azure.com/dan1wang/gts/_build/results?buildId=40
Circle CI: https://circleci.com/gh/dan1wang/gts/143


ofrobots:
@dan1wang thanks for the contribution. I wasn't able to find time today. I'll take a look at this PR tomorrow.

test/test-util.ts
```diff
@@ -141,9 +140,12 @@ describe('util', () => {
       include: ['c'],
     };
     const myMap = new Map();
-    myMap.set('/some/fake/directory/tsconfig.json', FAKE_CONFIG1);
-    myMap.set('/some/fake/directory/foo/FAKE_CONFIG2', FAKE_CONFIG2);
-    myMap.set('/some/fake/directory/foo/bar/FAKE_CONFIG3', FAKE_CONFIG3);
+    myMap.set(PATH_TO_TSCONFIG, FAKE_CONFIG1);
+    myMap.set(path.resolve(FAKE_DIRECTORY, './foo/FAKE_CONFIG2'), FAKE_CONFIG2);
```
soldair: using path.resolve here on win32 will flip all the slashes. is this a problem?
see 
`path.win32.resolve('/a/b/c','./foo/boo.js')` vs `path.posix.resolve('/a/b/c','./foo/boo.js')`

soldair: i see. we pass all of the path strings through resolve in some way before using them :+1:

PR #315, opened 2019-04-15
deps: switch to @npm/types for package.json types

ofrobots:
We get 'official' types from the horse's mouth now. Let's use those.
Oh, and `package-json` stopped exporting the types for package.json
itself. Let's not use that anymore.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -119,6 +119,12 @@
         "to-fast-properties": "^2.0.0"
       }
     },
+    "@npm/types": {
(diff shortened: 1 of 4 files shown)


JustinBeckwith:
One question here - what's with the `deps` prefix?  New to that one :)

ofrobots:
> One question here - what's with the deps prefix? New to that one :)

Good catch. I intended to say 'chore(deps):'

PR #318, opened 2019-04-19
Feature (closes #312) : template code at init

### LINKED ISSUE #312: TS18003: No inputs were found in config file
### Problem
tsc doesn't find anything to compile in the directories included in tsconfig.json.

### Repro steps

-  ```npx gts init```

- ```npm i```

### Workarounds

- delete the **prepare** script in package.json (this will prevent to compile on npm i)

OR

-  ```mkdir src && touch src/dummy.ts```

### Final thoughts
This behavior does not appear friendly to the user, better approach will be to include some dummy code to avoid the error.
Even better will be to generate some template code to get the new project up and running from the beginning and assure that everything is installed correctly.

### Proposition
- init.ts : new  function generateTemplateCode() will create src dir and an example index.ts,
- init.ts : modify function init() to call also generateTemplateCode(),
- package.json: create a **run** script that will compile and execute the template project.

pliski:
#### Description

At project initialization the src directory is created if it does not exist and  a  template is copied if  the src directory does not contain ts files already.

#### Todos:

- The template is just a placeholder for now, need of something fancier and more explanatory of gts.
- Waiting ofrobots/inline-fixtures#1 to add more tests.
- ```util.utilLog()``` logic should be in ```cli.Logger{}```.


pliski:
closing this one cause I'm preparing another one. Added more small features, fixes, etc

PR #320, opened 2019-04-22
Feature (closes #312) : template code at init

### LINKED ISSUE #312: TS18003: No inputs were found in config file

pliski:
At project initialization the src directory is created if it does not exist and a template is copied if the src directory does not contain ts files already.


src/util.ts
```diff
@@ -140,3 +153,84 @@ export function isYarnUsed(existsSync = fs.existsSync): boolean {
...
+ * If already exist check if contains any ts file.
+ * If it returns true we can expect that the directory
+ * exist and does not contain ts files.
+ */
+export async function createSrcDir(
+  dirpath: string,
+  options?: Options
+): Promise<boolean> {
+  try {
+    fs.mkdirSync(dirpath, { mode: 0o774 });
```
ofrobots: is there a reason to pass in the mode? This may end up being more permissive than what the user wanted. I suspect that the Node.js core official docs are incorrect - they claim the default to be `0o777`. That's not the actual behavior though:

image

src/util.ts
```diff
@@ -140,3 +153,84 @@ export function isYarnUsed(existsSync = fs.existsSync): boolean {
...
+        (file: string) => path.extname(file).toLowerCase() === extension
+      );
+      if (firstTsFileName !== undefined) {
+        return true;
+      }
+    }
+    return false;
+  } catch (error) {
+    utilLog(error, options);
+    return false;
```
ofrobots: Would it be safer (conservative) to return true here? Personally I think it would be better to let the error bubble to the caller. This seems like a simple boolean check, and it doesn't seem to have enough context to know what to do with an error.

test/test-init.ts
```diff
@@ -34,6 +35,7 @@ const OPTIONS_YES = Object.assign({}, OPTIONS, { yes: true });
 const OPTIONS_NO = Object.assign({}, OPTIONS, { no: true });
 const OPTIONS_YARN = Object.assign({}, OPTIONS_YES, { yarn: true });
 const MINIMAL_PACKAGE_JSON = { name: 'name', version: 'v1.1.1' };
+// const OPTIONS_LOG = Object.assign({}, OPTIONS_YES, { logger: console });
```
ofrobots:
```suggestion
```

src/util.ts
pliski: That's right the official doc is out of sync, I just wanted to avoid the 0777 but leave the default behavior is better.

src/util.ts
pliski: The aim of the try-catch here is : 

1) declutter createSrcDir() which is already in catch at this point;
2) catch readdirSync() exceptions.

#### 1) declutter createSrcDir()

If you think it is cleaner I could create an intermediate function, like:
```typescript
function createSrcDir(...){
    try {
        fs.mkdirSync(dirpath);
        return true;
     } catch (error) {
       if (error.code === 'EEXIST') {
          return srcAlreadyPresent(dirpath, options);
      }
    return false;
  }
}

function srcAlreadyPresent(...){
  try {
      return (await isAnyTsFileInDir(dirpath));
  } catch (error) {
    return false;
  }
}

function isAnyTsFileInDir() {
  // ... some logic here
}
```

#### 2) catch readdirSync exceptions.

For Example a directory exist but we don't owe the right permissions.
The solution at 1) will avoid the try-catch here, and we could safely bubble up the exceptions.
My though were that the installation of the template is an optional step so I tried to make as silent as I could, to avoid to block the process for an unwanted feature, for instance when adding gts to an existing project.
Please tell me if this make sense or if you prefer another behavior.
As a side note this is the case that I would like to address with furthers test as in ofrobots/inline-fixtures#1

pliski:
@ofrobots PR now include inline-fixtures@1.1.0 and is ready to be merged.

ofrobots:
This was quite entangled with merge conflicts, and I wanted to make code changes/reorganizations on top of your commits (rather than asking you to do it). Moving over to #340.

PR #321, opened 2019-04-23
build: test on node 12

ofrobots:


ofrobots:
Need to await for circleci to have a node:12 docker image available.

ofrobots:
TBR= @google/google-node-team

PR #332, opened 2019-05-03
build: Configure Cirrus CI for Windows

fkorotkov:
I saw #331 and decided to take a look. 

@JustinBeckwith @ofrobots PTAL

PR #333, opened 2019-05-04
build: use cirrus for CI

JustinBeckwith:
#332 added Windows testing for CirrusCI, now let's move the rest of our CI from Circle over.


fkorotkov:
For Windows stuff you can check #332. 😅

.cirrus.yml
```diff
@@ -9,3 +11,37 @@ task:
...
+    - npm run codecov
+
+publish_task:
+  only_if: $CIRRUS_TAG =~ '^v.*'
+  depends_on:
+    - test_task
+    - lint_task
+    - windows_task
+  environment:
+    NPM_TOKEN: ENCRYPTED[188593d2219e49e28e05c983a768a7e89ef89441f07f450a58744161674b2e7f127caff0a8021bb5bb75a09d4e18750f]
```
ofrobots: Does this have a race where both Circle and Cirrus would race to do a publish?

JustinBeckwith: This PR deletes the CircleCI config... so I don't think so

.cirrus.yml
```diff
@@ -9,3 +11,37 @@ task:
...
+
+coverage_task:
+  script:
+    - npm ci
+    - npm test
+    - npm run codecov
+
+publish_task:
+  only_if: $CIRRUS_TAG =~ '^v.*'
+  depends_on:
```
fkorotkov: You don't need `_task` suffix. Please remove it. ;-)

fkorotkov: Fixed in #345

PR #335, opened 2019-05-09
refactor: drop pify, and select dev dependencies

JustinBeckwith:
This drops a few dependencies:
- Move to `util.promisify` instead of `pify` now that we dropped node 6
- Use `fs-extra`, which was already included, instead of `npc`
- Use `execSync` instead of `execa` 
- Drops unused dependencies and type packages


JustinBeckwith:
@ofrobots is that windows test normally uh, passing?

ofrobots:
@soldair was poking at it last, and I think got it to pass.

soldair:
im pretty sure we did in fact get it to pass. i'll look into what happened with that

soldair:
https://github.com/google/gts/pull/330

JustinBeckwith:
I believe so.  In that case, I'll come back around and get this on #333

PR #340, opened 2019-05-11
feat: install a default src template

ofrobots:
A template is copied in the project if the src directory is not
present or does not contain ts files.

Fixes: https://github.com/google/gts/issues/312

Supercedes: #320


ofrobots:
Okay.. I wonder if I am running into a Cirrus bug here.

Trying to access a directory with permissions `0o000` is expected to throw. On my local machine and on CircleCI, it does. On Cirrus it doesn't.

PR #342, opened 2019-05-11
test: get windows tests passing

JustinBeckwith:
The special behavior for windows and npx wasn't needed, and things just run a little slower here :)  These seem to pass running on my windows VM.  Eager to see what happens over in #332!

### DIFF
--- a/test/mocha.opts
+++ b/test/mocha.opts
@@ -1,3 +1,3 @@
 --require source-map-support/register
---timeout 40000
+--timeout 160000
 --throw-deprecation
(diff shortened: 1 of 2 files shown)


ofrobots:
Waiting for the tests to pass before approving.

JustinBeckwith:
So hilariously, that probably won't happen in this iteration.  The tests are passing in a Windows VM when I run them locally.  In the kokoro Windows VM I am seeing:

```
T:\src\github\gts>call npm install node@8.16.0   || goto :error
'npm' is not recognized as an internal or external command,
operable program or batch file.
T:\src\github\gts>exit /b 1
```

I think there's a bug with the VM image we have in kokoro where it doesn't seem to be adding `npm` to the path? 

I know this is annoying, but I'd like to get this in, and validate it works with #332

PR #345, opened 2019-05-13
build: use correct task names in the CI config

fkorotkov:
Fixed task names


JustinBeckwith:
Yikes!  Good catch, and thanks for the fix.  Side note - maybe the CI run should fail if you depend on a task that technically doesn't exist :)

fkorotkov:
@JustinBeckwith it used to be enforcing it too much to the point that if there was `only_if` statement that removed a task it was reported as missing. There should be something in between to see if a task from dependencies is missing for a reason. 🤔

PR #369, opened 2019-07-30
fix: process less files in parallel preventing OOM errors

bcoe:
## The Problem

tslint seems to have a memory leak such that, in codebases with a lot of files like `google-api-nodejs-client`, we're bumping into OOM issues, here's an example: 

https://github.com/googleapis/google-api-nodejs-client/pull/1778

The problem crops up when `getResult` is called on a large set of files.

## Solution

It doesn't seem like, for the purposes of linting, there should be any state shared between the files added to the linter, this PR proposes linting a maximum of 25 files in conjunction, and then creating a new `Linter` instance.

I've confirmed that this fixes our issues with `google-api-nodejs-client`.

### DIFF
--- a/src/lint.ts
+++ b/src/lint.ts
@@ -20,8 +20,7 @@ import * as ts from 'typescript';
 
 import { Options } from './cli';
 
-// how many files should we lint in conjunction before creating a new linter?
-const MAX_FILES_LINTED = 25;
+const MAX_PARALLEL_FILES = 25;
 
 /**
  * Run tslint with the default configuration. Returns true on success.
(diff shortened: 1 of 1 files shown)


ofrobots:
Would it be better to create a new `Linter` for each file? or does that do a lot redundant TypeScript reparsing?

src/lint.ts
```diff
@@ -74,19 +76,36 @@ export function lint(
...
+        if (linter) {
+          const result = linter!.getResult();
+          if (result.errorCount || result.warningCount) {
+            options.logger.log(result.output);
+            return false;
+          }
+        }
+        linter = new Linter({ fix, formatter: 'codeFrame' }, program);
+      }
+      const file = files[i];
```
JustinBeckwith: If there are say 27 files - doesn't this mean we'd get to this code block without having instantiated the linter?

bcoe: I don't think so, because `0 % 27` is `0`, so we'll instantiate the linter. Might be worth trying @ofrobots approach of just creating a linter each time, let me try.

JustinBeckwith: 🤦‍♂ I don't know why my brain read the loop as going down instead of up.  Carry on!

bcoe: It's okay, I think @ofrobots's suggestion of simplifying to creating a linter each time is better any ways.

src/lint.ts
```diff
@@ -74,19 +74,20 @@ export function lint(
       : path.resolve(options.gtsRootDir, 'tslint.json');
 
     const configuration = Configuration.loadConfigurationFromPath(configPath);
-    const linter = new Linter({ fix, formatter: 'codeFrame' }, program);
 
-    files.forEach(file => {
+    for (let i = 0; i < files.length; i++) {
```
ofrobots: minor style nit: Implicit iteration via `for .. of` loop would be more readable.

PR #400, opened 2019-11-26
feat!: use eslint instead of tslint

JustinBeckwith:
BREAKING CHANGE: Switch from tslint to eslint.  This will impact all lint rules, and tslint specific configuration.


src/cli.ts
```diff
@@ -105,15 +109,30 @@ async function run(verb: string, files: string[]): Promise<boolean> {
...
+  }
+
   switch (verb) {
-    case 'check':
-      const passLint = await lint(options, files);
-      const passFormat = await format(options, files);
-      return passLint && passFormat;
+    case 'check': {
+      try {
+        await execa('npx', ['eslint', ...flags], { stdio: 'inherit' });
```
JustinBeckwith: @bcoe @google/google-node-team so I'm interested in feedback on this one, because I'm a bit torn.  I'm experimenting with getting rid of the wrapping invocation of `prettier` and `tslint`, and instead just shelling out to `eslint`.  I'm doing this for a few reasons:
- With `tslint` and `prettier`, we had to run the format and lint command separately.  With `eslint-prettier`, we can combine them into a single call now.
- The fewer moving parts in this library the better.  Since `eslint` pretty much does what we want, just shelling out seems like a lot less can go wrong.  

That having been said - I recognize there are probably ways this could go south.  Specifically testing it will be much harder 😆   What am I missing here?  What do others think?

PR #428, opened 2020-01-13
build: enable publication

bcoe:
We also need to [turn the bot on](https://github.com/apps/publish-bot) for this repo.


bcoe:
Oh yeah, I wasn't aware of this. I would consider doing so personally, in the name of standardization.

JustinBeckwith:
Could I trouble you to remove it in this PR?  Thinking if this bot does the `npm publish`, we probably should remove it before we enable the bot.

PR #439, opened 2020-02-03
build: use github actions

JustinBeckwith:


.github/workflows/ci.yaml
```diff
@@ -0,0 +1,66 @@
...
+      - uses: actions/checkout@v1
+      - uses: actions/setup-node@v1
+        with:
+          node-version: 12
+      - run: npm ci
+      - run: npm test
+      - run: npm run codecov
+      - uses: codecov/codecov-action@v1
+        with:
+          token: ${{ secrets.CODECOV_TOKEN }}
```
bcoe: I noticed on wombat dressing room permission issues on the upload step of codecov, we should check the logs for the action when this runs.

JustinBeckwith: I had to manually set the `CODECOV_TOKEN`, which is something we'd need to figure out if looking at this broadly for client libs.  Other than that - it works!

PR #452, opened 2020-03-04
refactor: compile using tsc with no arguments

Maxim-Mazurok:
Since TypeScript [v1.5.0-alpha](https://github.com/microsoft/TypeScript/blob/b18dd53242df31163390c411ac7ab5e9c282fea2/src/compiler/tsc.ts#L185) `tsc` will search for `tsconfig.json` in `cwd` by default.

First implemented in https://github.com/microsoft/TypeScript/pull/1692.

So, there's no need in specifying project path anymore and we can safely replace `tsc -p .` with just `tsc`.

### DIFF
--- a/package.json
+++ b/package.json
@@ -20,7 +20,7 @@
     "build": "npm run compile",
     "clean": "rimraf ./build/",
     "codecov": "c8 report --reporter=json && codecov -f coverage/*.json",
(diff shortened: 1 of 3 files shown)


bcoe:
@Maxim-Mazurok thank you for the contribution 👌

PR #457, opened 2020-03-19
fix(deps): address minor security warning with minimist

bcoe:


bcoe:
nvm, this isn't actually patched in eslint yet https://github.com/eslint/eslint/issues/13050

PR #487, opened 2020-04-12
feat!(eslint): reinstate variable-name as naming-convention #486

### LINKED ISSUE #486: variable-name removed in gts@2.0.0
TSLint's `variable-name` rule existed in `gts@1`:

https://github.com/google/gts/blob/4f211518984bd09a346624f86167c4da965627b5/tslint-rules.json#L70-L76

Per [typescript-eslint's ROADMAP style section](https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/ROADMAP.md#style) (migration guide?), the TSLint `variable-name` equivalent is now `@typescript/naming-convention`. See [rules/naming-convention](https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/naming-convention.md).

Since `variable-name` was removed and `naming-convention` was not included, variable name linting is now more relaxed, i.e. a possible regression.

While the new `naming-convention` allows for significantly more control, a 1:1 replacement would be:
```
---
rules:
  '@typescript-eslint/naming-convention':
    - error
    - selector: variable
      format: ['camelCase', 'UPPER_CASE']
      leadingUnderscore: allow
      trailingUnderscore: allow
      custom:
        regex: 'any|[Nn]umber|[Ss]tring|[Bb]oolean|[Uu]ndefined'
        match: false
```

`naming-convention` does not explicitly include support for the `ban-keywords` option, so `custom` with a regex was included above to provide equivalency. See https://github.com/palantir/tslint/blob/8efa78e35a88b62775ce3c9bf0fa00586b0d34e1/src/rules/variableNameRule.ts#L26-L36

zamnuts:
BREAKING CHANGE

- configured as `error` level
- not 1:1 with legacy `variable-name`, does not blacklist primitive type names in identifier name
- scoped to _variables_ only, e.g. no effect on interfaces for method names etc. see https://github.com/google/gts/pull/487/files#r410749633

Fixes #486 

~Configured as `warn` level so as not to be a breaking change, unless using `--max-warnings`.~

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -34,7 +34,7 @@
         "@typescript-eslint/explicit-function-return-type": "off",
         "@typescript-eslint/camelcase": "off",
         "@typescript-eslint/naming-convention": [
-          "warn",
+          "error",
           {
             "selector": "variable",
             "format": ["camelCase", "UPPER_CASE"],
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -33,6 +33,19 @@
         "@typescript-eslint/no-empty-function": "off",
         "@typescript-eslint/explicit-function-return-type": "off",
         "@typescript-eslint/camelcase": "off",
+        "@typescript-eslint/naming-convention": [
+          "warn",
+          {
+            "selector": "variable",
+            "format": ["camelCase", "UPPER_CASE"],
+            "leadingUnderscore": "allow",
+            "trailingUnderscore": "allow",
+            "custom": {
+              "regex": "any|[Nn]umber|[Ss]tring|[Bb]oolean|[Uu]ndefined",
+              "match": false
+            }
+          }
+        ],
         "node/no-missing-import": "off",
         "node/no-empty-function": "off",
         "node/no-unsupported-features/es-syntax": "off",
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -39,11 +39,7 @@
             "selector": "variable",
             "format": ["camelCase", "UPPER_CASE"],
             "leadingUnderscore": "allow",
-            "trailingUnderscore": "allow",
-            "custom": {
-              "regex": "any|[Nn]umber|[Ss]tring|[Bb]oolean|[Uu]ndefined",
-              "match": false
-            }
+            "trailingUnderscore": "allow"
           }
         ],
         "node/no-missing-import": "off",


.eslintrc.json
```diff
@@ -33,6 +33,19 @@
...
         "@typescript-eslint/camelcase": "off",
+        "@typescript-eslint/naming-convention": [
+          "warn",
+          {
+            "selector": "variable",
+            "format": ["camelCase", "UPPER_CASE"],
+            "leadingUnderscore": "allow",
+            "trailingUnderscore": "allow",
+            "custom": {
+              "regex": "any|[Nn]umber|[Ss]tring|[Bb]oolean|[Uu]ndefined",
```
bcoe, line 44: I'm a little worried about this list of exceptions growing unbounded, is there a configuration setting that we could use that just captures common constants?

.eslintrc.json
```diff
@@ -33,6 +33,19 @@
         "@typescript-eslint/no-empty-function": "off",
         "@typescript-eslint/explicit-function-return-type": "off",
         "@typescript-eslint/camelcase": "off",
+        "@typescript-eslint/naming-convention": [
+          "warn",
+          {
+            "selector": "variable",
+            "format": ["camelCase", "UPPER_CASE"],
```
bcoe, line 40: For interfaces, I believe we use the format:

```
MyInterface
```

Which wouldn't match the camel case or all caps regex?

JustinBeckwith:
So....  This would kinda be a breaking change right?

.eslintrc.json
zamnuts: The `naming-convention` plugin is aware of ECMAScript IdentifierName contexts using the `selector` property. In this case, the format rule is only applied to Identifiers that are `variable`s. Other selectors exist such as `function`, `method`, `class`, `typeAlias`, and for the case you describe, `interface`. Therefore, this format will not apply to interface IdentifierNames such as `MyInterface`.

I'd be happy to add more `naming-convention` rules, especially since I am of the opinion that we should be a bit more strict. See https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/naming-convention.md#individual-selectors

I actually tested this against nodejs-bigquery and found only 3 linting problems related to this change, but only because changes were made to accommodate gts@2.

zamnuts:
> So.... This would kinda be a breaking change right?

@JustinBeckwith, yes, this is _technically_ breaking in the absolute sense, but only if consumers of `gts` used the `--max-warnings` option in `eslint`. Since this PR only classifies this rule as a `warn`, for most (?) cases beyond those that use `--max-warnings`, this will not break.

However, `gts` CLI usage invokes `eslint` under the hood with a fixed set of options, so I presume that the usage of `--max-warnings` does not exist.

.eslintrc.json
zamnuts: I was not a fan of this either, but wanted parity with the old `variable-name` rule.

The scope of this regex is reserved words. I looked, but was not able to find a configuration that had equivalency, or a dictionary of such words that could be used in this context.

This list was taken from a list of banned keywords in tslint's variable-name feature: https://github.com/palantir/tslint/blob/a66f3b3430b4d3f03ff93fcef94609b88809d4e1/src/rules/variableNameRule.ts#L26-L36

zamnuts: @bcoe I have remove the keyword blacklist regex

JustinBeckwith:
Having sat with the current functionality ... I don't think this change is worth the pain it would cause.  We may come back around at some point, but I just don't want to introduce the breaking change.

PR #499, opened 2020-05-04
fix: prohibit calls to `it.only` and `describe.only`

thebrianchen:
Prevent calls to ".only" in mocha tests. Porting upstream from nodeJS-firestore (original [PR](https://github.com/googleapis/nodejs-firestore/pull/1068)).


thebrianchen:
@cc beckwith

PR #501, opened 2020-05-06
fix: better error message for broken tsconfig.json

alexander-fenster:
If `tsconfig.json` cannot be parsed (invalid JSON) we should not say that it's not found.

### DIFF
--- a/src/util.ts
+++ b/src/util.ts
@@ -76,10 +76,6 @@ async function getBase(
   readFiles.add(filePath);
   try {
     const json = await customReadFilep(filePath, 'utf8');
-  } catch (err) {
-    throw new Error(`${filePath} not found`);
-  }
-  try {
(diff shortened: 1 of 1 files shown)


src/util.ts
```diff
@@ -91,7 +91,7 @@ async function getBase(
 
     return contents;
   } catch (err) {
-    throw new Error(`${filePath} Not Found`);
+    throw new Error(`Error: ${filePath}: ${err}`);
```
JustinBeckwith, line 94: So good catch - how about instead, we prepend the `${filePath} not found` to the original `err.message`, and then re-throw `err` so we maintain the full stack trace?

bcoe: could you instead capture the stack trace, so that the original stack is in the error, but throw the new message? it feels a little weird mucking with an error (a little too magic).

PR #518, opened 2020-05-26
use ES2020

naseemkullah:
Many goodies:

https://medium.com/@selvaganesh93/javascript-whats-new-in-ecmascript-2019-es2019-es10-35210c6e7f4b

https://www.digitalocean.com/community/tutorials/js-es2020

### DIFF
--- a/tsconfig-google.json
+++ b/tsconfig-google.json
@@ -4,7 +4,7 @@
     "allowUnusedLabels": false,
     "declaration": true,
     "forceConsistentCasingInFileNames": true,
-    "lib": ["es2018"],
+    "lib": ["es2020"],
     "module": "commonjs",
     "noEmitOnError": true,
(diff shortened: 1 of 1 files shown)


JustinBeckwith:
Greetings, and thanks for the PR!  Sadly, we can't target `es2020` just yet.  This library aims to support all LTS supported versions of node.js, as well as "latest":
https://nodejs.org/en/about/releases/

This includes (as of this moment) v10, v12, and v14. The `es2020` config includes a variety of features that aren't available yet in node.js 10.x.  For example:
https://node.green/#ESNEXT-candidate--stage-3--instance-class-fields-private-instance-class-fields-basic-support

The good news is that for consumers of this lib, you can easily override the target version in your `tsconfig.json` :)  Thanks!

naseemkullah:
Ok, thanks @JustinBeckwith

PR #521, opened 2020-05-28
feat: generate .eslintignore when running init

naseemkullah:
- generate `.eslintignore` and include `build/`

- add empty line at enf of generated `.prettierrc.js`

solves https://github.com/google/gts/issues/483

solves https://github.com/google/gts/issues/520

### DIFF
--- a/src/init.ts
+++ b/src/init.ts
@@ -245,7 +245,8 @@ async function generateTsConfig(options: Options): Promise<void> {
 async function generatePrettierConfig(options: Options): Promise<void> {
   const style = `module.exports = {
   ...require('gts/.prettierrc.json')
-}`;
+}
+`;
   return generateConfigFile(options, './.prettierrc.js', style);
 }
 
--- a/src/init.ts
+++ b/src/init.ts
@@ -181,6 +181,8 @@ export const ESLINT_CONFIG = {
   extends: './node_modules/gts/',
 };
 
+export const ESLINT_IGNORE = 'build/\n';
+
 async function generateConfigFile(
   options: Options,
   filename: string,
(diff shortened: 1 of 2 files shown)


bcoe:
@naseemkullah speaking with my colleague @JustinBeckwith, historically we populated a default `.eslintignore` with some reasonable values such as:

```
build/
docs/
```

Rather than adding this into configuration, what if we were to again generate this file during initialization?

naseemkullah:
sgtm @bcoe 

ooc is there any fundamental benefit to either approach?

bcoe:
@naseemkullah from my perspective, there tends to be a bit of variation in the sorts of files folks want to ignore on a project by project basis. _for instance, as Googlers, we have protos/ generated in a lot of our libraries._

The `.eslintignore` feels like it's less buried, making it clear where a consumer can easily change these rules.

naseemkullah:
Thanks @bcoe I tried implementing this but am unsure how to test locally, could you please advise?

bcoe:
@naseemkullah your test looks reasonable to me,  to test you can go:

```
npm link .
gts init
```

and see whether the file generates as expected.

naseemkullah:
> @naseemkullah your test looks reasonable to me, to test you can go:
> 
> ```
> npm link .
> gts init
> ```
> 
> and see whether the file generates as expected.

Thanks @bcoe, it appears that `gts` was already an alias for `git tag` for some reason, but 

```
npm link .
/usr/local/bin/gts init
```

did work and all is in order.

naseemkullah:
I took the liberty of including another small fix, the prettier config was not generating with an empty line at eof but now is.

bcoe:
@naseemkullah thank you for the contribution 👍

naseemkullah:
My pleasure @bcoe.

PR #528, opened 2020-06-09
fix(deps): update typescript-eslint monorepo to v3 (major)

JustinBeckwith:


alexander-fenster:
Unrelated to this change, I think it would make perfect sense to have a baseline-kind of tests for `gts` - a random code that gets formatted and compared to the checked in baseline. That way we would catch if a random `eslint` or config change triggers any incompatibility in formatting.

PR #537, opened 2020-07-15
feat: ignore 'no-unused-vars' for _ var prefix

bcoe:
It's a common convention to use a `_` to specify that you do not intend to use a variable, an example would be the following code:

```js
(_err, req) => {
   console.info(req.headers)
}
```

This enables this functionality.

Refs: https://github.com/GoogleCloudPlatform/wombat-dressing-room/pull/76


bcoe:
talking with @JustinBeckwith, perhaps this is configuration that people should set on a repo by repo basis.

PR #640, opened 2021-05-21
fix(deps): pin deps for "hermetic" builds

bcoe:
Changes to linting in minor versions of deps is making it hard to ship our linter within a docker container.


JustinBeckwith (changes requested):
Let's maybe talk this one through.

PR #653, opened 2021-07-12
Added binPrefix option, to support running via Bazel

dorfire:


richardbarrell-calvium:
I have another pull request with a similar purpose at #654. Is there any chance that the change I proposed is suitable for the use-case that you have, please?

dorfire:
Not entirely sure; Bazel is very particular about dependency paths

richardbarrell-calvium:
Rats. I was hoping that it would be easy for Bazel users to arrange to add the eslint dir to $PATH.

tonycoco:
I can help push this into a release if we can resolve the conflicts

tonycoco:
Not sure this change is necessary any longer. @dorfire can you comment on this being necessary still? Otherwise, will close out the PR as **out of date**.

loeffel-io:
+1

PR #686, opened 2022-05-26
build!: drop support for node.js 10.x

JustinBeckwith:


bcoe (approved):

bcoe:
@JustinBeckwith I believe we need to continue supporting Node 12 for the time being, while many of the client libraries continue to do so -- although I would like to gradually get closer to following the same timeline as Node.js.

PR #693, opened 2022-06-08
fix(deps): upgrade to eslint 8.x

JustinBeckwith:
Details on the funky license for `argparse` is covered here:
https://github.com/nodeca/argparse/issues/159

Let's uh, maybe get this stuff into node core so we don't have to deal with it 😛


alexander-fenster (approved):

alexander-fenster:
👋  @JustinBeckwith

PR #765, opened 2023-06-30
feat: move to eslint-plugin-n

aladdin-add:
eslint-plugin-n is a community-maintained fork of eslint-plugin-node: https://github.com/eslint-community/eslint-plugin-n

It ensures that an important module like eslint-plugin-node is kept up to date, with both ESLint 8.x and with Node.js itself.

Similar PR: electron/electron#38874

Full disclosure: I'm part of the ESLint Community Core Team.

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -1,11 +1,11 @@
 {
   "extends": [
     "eslint:recommended",
-    "plugin:node/recommended",
+    "plugin:n/recommended",
     "prettier"
   ],
   "plugins": [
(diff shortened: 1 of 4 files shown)


aladdin-add:
looks like the package still supports node.js v12, while eslint-plugin-n has dropped node.js < 16 (EOL). You can merge the PR after dropping node.js < 16. :)

tonycoco:
Going to track changes here: #865 with tests fixed and passing

PR #822, opened 2023-10-21
feat!: Update '.prettierrc.json' to include `trailingComma: "all"` to match internal Google config

### LINKED ISSUE #662: Enable more trailing commas, particularly in function parameters and calls
I'd like to suggest changing the Prettier trailingComma setting from es5 to all for the next gts release with breaking changes.

https://prettier.io/docs/en/options.html#trailing-commas

In particular, this would enable trailing commas in multiline function parameter lists and calls, which can help reduce commit diff noise.

The Prettier docs note that this requires ES2017 and TypeScript 2.7, but gts already specifies ES2018 and TypeScript 4.x.

calebegg:
fixes #662

### DIFF
--- a/.prettierrc.json
+++ b/.prettierrc.json
@@ -1,6 +1,6 @@
 {
   "bracketSpacing": false,
   "singleQuote": true,
-  "trailingComma": "es5",
+  "trailingComma": "all",
   "arrowParens": "avoid"
 }
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "gts",
-  "version": "5.2.0",
+  "version": "6.0.0",
   "description": "Google TypeScript Style",
   "repository": "google/gts",
   "main": "build/src/index.js",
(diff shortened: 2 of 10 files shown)


tonycoco:
Can we rebase this? Change looks good to me.

tonycoco (approved):
LGTM

calebegg:
I am not good at git, my brain has been permanently disfigured by fifteen years of using nothing but Piper/Perforce. I think I screwed this up somehow even though I was following a github help page. If you want to help me that's fine, or if you want to just make the change yourself that's fine too. I've spent half an hour on this and I'm quite frustrated.

calebegg:
Ok I think I fixed it (thanks again for the help!)

tonycoco (approved):

tonycoco:
Note: the commit message needs to follow conventional commits. Can you amend your commit message?

This change will be messy for lots of repos using `gts` so we should think about bumping the release semver to align.

calebegg:
Done

PR #866, opened 2024-03-22
build: test on node 14, 16, 18, 20, and 21

tonycoco:
The End-of-Life date for Node.js 16 was [moved forward by seven months](https://nodejs.org/en/blog/announcements/nodejs16-eol/) to coincide with the end of support of OpenSSL 1.1.1 on September 11th, 2023.

This fixes all warnings on Node 21 as well as removes the unnecessary type defs packages.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -1,6 +1,6 @@
 on:
   push:
-    branches: [ $default-branch ]
+    branches: [$default-branch]
   pull_request:
 name: ci
 jobs:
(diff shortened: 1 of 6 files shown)


alexander-fenster:
Let's make it a breaking change. We still support `>=14` in Cloud client libraries.

tonycoco:
Amended my commit to include Node 14, 16, 18, 20, and 21.

alexander-fenster (approved):

PR #962, opened 2026-02-28
chore(deps): replace chalk with picocolors

roli-lpci:
## Summary

- Replace `chalk` (4 transitive dependencies: chalk, ansi-styles, supports-color, has-flag) with `picocolors` (zero dependencies) for terminal color output
- All usage is single-method calls (`bold`, `red`, `green`, `gray`, `blue`) which picocolors supports directly
- Reduces production dependency tree surface area

## Changes

- `src/init.ts` — import and 6 call sites updated
- `src/clean.ts` — import and 5 call sites updated
- `test/kitchen.ts` — import and 1 call site updated
- `package.json` — swap chalk for picocolors in dependencies
- `package-lock.json` — regenerated

## Testing

- `npm run compile` — clean
- `npm run lint` — clean
- `npm test` — 32/32 passing, 100% function coverage


roli-lpci:
@googlebot I signed it.

tonycoco:
I quite like this change to reduce deps. Will review when I get a minute. Thanks!

